PR #21, opened 2020-07-01
#20 map rest dgraph queries 

### LINKED ISSUE #20: Map REST queries to Dgraph queries
Add logic for mapping v1 HTTP requests to graphql queries (from #9 )

wlthomson:
Fixes #20 (and lots of minor tidy ups that I should have split into different PRs, but doesn't seem worth the effort at this stage!).

- Updates dgraph test data
- Adds dgraph cleanup script
- Updates query mappings
- API tidy ups and formatting
- Bumps `package-lock.json`


api/.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
wlthomson, line 1: Should probably add this to `.gitignore`?

api/yarn.lock
```diff
@@ -0,0 +1,5310 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
wlthomson, line 1: Looks like some of us are using `yarn` and some `npm`. I started off with `npm`, heard it's better than `yarn` nowadays. Thoughts?

dgraph/sample/data.json
```diff
@@ -14,175 +14,12 @@
           "value": "6.2.1 Beta-lactam medicines"
         },
         {
-          "description": "Oral dose forms",
```
wlthomson, line 17: Think there might be some confusion between us on the schema? Have updated to what I think is the best/easiest at this stage. Feedback welcome :).

api/yarn.lock
mark-prins: ✋  guilty. `npm` is old school, when `yarn` came along, it was new and shiny so I still think of it that way. 
Both work about the same.. so I don't mind which we use. I'll stop yarning.

api/src/mappers.js
```diff
@@ -0,0 +1,94 @@
...
+    },
+    { parameters: { valid: [], invalid: [] } }
+  );
+};
+
+/**
+ * Map API request to GraphQL+- payload. Query parameters are assumed valid.
+ *
+ * @param  {Object} request REST api request.
+ * @return {String}         GraphQL+- response from Dgraph server.
```
katherine-sussol, line 31:
```suggestion
 * @return {String}         GraphQL+- request payload for Dgraph server.
```
Super nit-pick at the comment - at this point it's still a request payload, I think?

api/src/mappers.js
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }`;
+  }
+};
+
+/**
+ *
+ * Map Dgraph response to valid v1 API JSON payload.
+ *
+ * @param  {Object} request  GraphQL+- request sent to Dgraph server.
```
katherine-sussol, line 67:
```suggestion

```
Another comment thing - I think your description/naming is definitely better than my original one though ;)

wlthomson: Oops. 20940fe6a9416424f8d8ad774e84d7722dbbc614.

api/src/mappers.js
wlthomson: Good point. 20940fe6a9416424f8d8ad774e84d7722dbbc614.

dgraph/sample/data.json
katherine-sussol: I like this updated schema so far 👍 . (The two of us did discuss most of it in person but for posterity and to make sure I am on the same page 😄 )

Think these were the main schema/structure changes and rationale?
- form_category changed from a property to child relation 
  - Made going straight up/down the tree much easier (can recursively go down children to get to level of most detail rather than having to go via a property to get to a child) 
- Appending additional detail to description as we get further down the 'tree'
  - Dgraph description searching/building is now much easier (previously we had to build up the description string as we went down the tree on the client side, now we can just query for the specific node we want and use the description as is)
- Synonyms removed - problem for later? (Maybe to be added as properties?)

api/.vscode/launch.json
mark-prins: I thought it might be useful for others, as long as people are using vscode, of course.
Just an example of debugging launch script. Happy to exclude though if you think its not useful.

dgraph/sample/data.json
mark-prins: > Synonyms removed

hm? They were in the wiki example, which is why I had them in. Similarly some of the other changes have now digressed from the wiki example. is there a reason for doing that?

api/src/mappers.js
```diff
@@ -0,0 +1,93 @@
...
+ * ]
+ */
+export const mapResponse = (response) => {
+  const { data } = response;
+  if (data.length - 1) {
+    return JSON.stringify(
+      data?.query.map(({ code, description }) => ({ code, name: description })) ?? []
+    );
+  } else {
+    const [{ code, description }] = data?.query ?? [];
```
katherine-sussol, line 90: This line is currently throwing an error when there are no results 😿 . I'm still trying to get my head around that left hand side syntax (is it trying to create a of object from data.query.code and data.query.description for each item in data.query?) - but it looks like data.query is an empty array in this case which is making it unhappy.

dgraph/sample/data.json
katherine-sussol: I think the removal of the Amox examples was an unintended side effect while restructuring - I'm happy to add those back in as a next step. I suppose it would also be good to get the rdf and json aligned as well?

Also think we should add the synonyms added back in - but not sure if we had a consensus on where they should go - do we still want them as properties of medicinal products?

wlthomson: Personally, I think they should be properties. For the current schema, properties can't be nested, so if we want to had children off synonyms, they need to be entities (but IMO that will make things confusing).

PR #23, opened 2020-07-02
#22 tidy up response mapper

### LINKED ISSUE #22: Error in response mapper
With reference to [this comment](https://github.com/openmsupply/unified-codes/pull/21#discussion_r448649832)

When no results returned there's an error in the `mapper.mapResponse`

mark-prins:
Fixes #22 

The `query` property of `data` is the array, not `data` itself. 
Have tweaked the mapper to reflect this and cope with an empty array and an undefined `query` or `data`

### DIFF
--- a/api/src/mappers.js
+++ b/api/src/mappers.js
@@ -82,12 +82,7 @@ export const mapRequest = (request) => {
  */
 export const mapResponse = (response) => {
   const { data } = response;
-  if (data.length - 1) {
-    return JSON.stringify(
-      data?.query.map(({ code, description }) => ({ code, name: description })) ?? []
-    );
-  } else {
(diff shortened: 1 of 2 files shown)


mark-prins:
> The problem here, I think, is the default case: [], which should be [{}].

Except that I tried that first - and that didn't work for me!

actually, I see why now. The query won't be undefined, it is an empty array. so the default case doesn't come into play!

I had some other concerns about the original implementation and confusion about intention. 
I think I get the intention now. will tweak.

mark-prins:
I've reworked the mapper to cater for
- no results (i.e. empty array)
- no data and/or no query
- return a single object if there is only one result

offroad: had to add some of the sample data back in, so that I could test multiple results.

PR #24, opened 2020-07-02
#14 expand rest api test data

### LINKED ISSUE #14: Expand rest api test data
### Description
- Add more mock graphql data to enable better testing of rest api

### Goal
Expand test data in order to be able to develop more comprehensive tests

katherine-sussol:
Fixes #14
 
Re-populated Amox example from [wiki](https://sussol.net/wiki/doku.php/msupply:planning:projects:universal_codes_v2) into sample data files.
- Added SnoMED CT and Drug Bank Url
- Added nested property to schema (the existing examples already had this, though it wasn't explicitly defined in schema)
- Align RDF with JSON

Possible next steps if all good with this schema:
- Add Paracetamol example to RDF
- More products
- More API test cases
- Add sample expected API results based on test data

### DIFF
--- a/dgraph/sample/data.json
+++ b/dgraph/sample/data.json
@@ -15,7 +15,13 @@
         },
         {
           "type": "drugbank_code",
-          "value": "DB01060"
+          "value": "DB01060",
+          "has_property": [
+            {
+              "type": "source",
+              "value": "https://drugbank.ca/drugs/" 
+            }
+          ]
         },
         {
           "type": "atc_code",
@@ -50,6 +56,12 @@
         {
           "description": "Amoxicillin oral",
           "type": "form_category",
+          "has_property": [
+            {
+                "value": "350162003",
+                "type": "SnoMED CT"
+            }
+          ],
           "has_child": [
             {
               "description": "Amoxicillin oral capsule",
--- a/dgraph/sample/data.json
+++ b/dgraph/sample/data.json
@@ -54,6 +54,7 @@
       ],
       "has_child": [
         {
+          "code": "AMOX0R47",
           "description": "Amoxicillin oral",
           "type": "form_category",
           "has_property": [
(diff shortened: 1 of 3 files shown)


dgraph/sample/data.json
```diff
@@ -14,23 +14,219 @@
...
+              "value": "544Y3K6MYH"
+            }
+          ]
         }
       ],
-      "has_child": []
+      "has_child": [
+        {
+          "description": "Amoxicillin oral",
+          "type": "form_category",
```
wlthomson, line 58: Wondering if we should add an arbitrary code here? Think would be tidier if all entity nodes had a `code`?

dgraph/sample/data.json
```diff
@@ -14,23 +14,219 @@
...
+          "has_property": [
+            {
+                "value": "350162003",
+                "type": "SnoMED CT"
+            }
+          ],
+          "has_child": [
+            {
+              "description": "Amoxicillin oral capsule",
+              "type": "form",
```
wlthomson, line 68: Same as above :).

dgraph/sample/data.json
katherine-sussol: Yeah makes sense - I guess we're still a bit unsure exactly which will and won't have codes, but we'd definitely want to make sure we can handle them if they exist.

Added in 622f2d4724afccb5ae04556103a8e5b003b21c90

dgraph/sample/data.json
katherine-sussol: Added in 622f2d4724afccb5ae04556103a8e5b003b21c90 (code choice may not have been completely arbitrary) 😸

mark-prins: I'd recently been made aware of how the current codes are generated: https://wiki.sussol.net/doku.php/projects:universal_drug_codes:accessing_rds_instance?s[]=universal&s[]=codes#add_a_code_new_universal_code

It's a substring of the UUID used in mSupply for the item. Quite likely not helpful here, as these will be new nodes. Just mentioning because I thought it interesting.

PR #26, opened 2020-07-03
#25 Update mapper.js

### LINKED ISSUE #25: Items endpoint failing invalid params test
As per screenshot below:
Screen Shot 2020-07-03 at 10 29 57 AM

UPDATE post-fix: 
mapper.js had some minor issues which meant the request would not get blocked:
- request.params is null in function (changed to request.query)
- request.url is null in function (changed to request.raw.url)
- full url with query string wasn't matching key to access PARAMETERS object
- reduce function was returning strange results e.g:
  - items?invalid=data&test=moredata => { valid: [], invalid: { query: 'test' } } 
  - updated to do simple filter of params on whitelist

katherine-sussol:
Fixes #25 

Found a few minor issues:
- request.params is null in function (changed to request.query)
- request.url is null in function (changed to request.raw.url)
- full url with query string wasn't matching key to access PARAMETERS object
- reduce function was returning strange results e.g.:
  - items?invalid=data&test=moredata => { valid: [], invalid: { query: 'test' } } 
  - Changed to filter on whitelist - simpler but not as fancy (could see what it was trying to do - for educational purposes may test it locally some more)

Happy to hear if any of the changes could be done better!

### DIFF
--- a/api/src/mappers.js
+++ b/api/src/mappers.js
@@ -11,7 +11,8 @@ const PARAMETERS = {
  */
 export const parseRequest = (request) => {
   const { url } = request.raw;
-  const whitelist = PARAMETERS[url.split('?')[0]];
+  const [baseUrl] = url.split('?');
+  const whitelist = PARAMETERS[baseUrl];
   const { query } = request;
 
   return {
--- a/api/src/mappers.js
+++ b/api/src/mappers.js
@@ -10,18 +10,16 @@ const PARAMETERS = {
  * @return {Object}                  Object containing valid and invalid parameters.
  */
 export const parseRequest = (request) => {
-  const { url } = request;
-  const whitelist = PARAMETERS[url];
-  const { params } = request;
-  return Object.keys(params).reduce(
-    ({ parameters }, param) => {
-      const { valid: validParameters, invalid: invalidParameters } = parameters;
-      if (whitelist.includes(param))
-        return { parameters: { valid: { ...validParameters, param }, invalid: invalidParameters } };
-      return { parameters: { valid: validParameters, invalid: { ...invalidParameters, param } } };
+  const { url } = request.raw;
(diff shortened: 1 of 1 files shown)


api/src/mappers.js
```diff
@@ -10,18 +10,16 @@ const PARAMETERS = {
...
-  const whitelist = PARAMETERS[url];
-  const { params } = request;
-  return Object.keys(params).reduce(
-    ({ parameters }, param) => {
-      const { valid: validParameters, invalid: invalidParameters } = parameters;
-      if (whitelist.includes(param))
-        return { parameters: { valid: { ...validParameters, param }, invalid: invalidParameters } };
-      return { parameters: { valid: validParameters, invalid: { ...invalidParameters, param } } };
+  const { url } = request.raw;
+  const whitelist = PARAMETERS[url.split('?')[0]];
```
wlthomson, line 14: Nit: is there a way to get the base url (i.e. without splitting on the query param)?

wlthomson, line 14:
```suggestion
  const [baseUrl] = url.split('?');
  const whitelist = PARAMETERS[baseUrl];
```

api/src/mappers.js
```diff
@@ -10,18 +10,16 @@ const PARAMETERS = {
...
+  const { query } = request;
+
+  return {
```
wlthomson, line 17: Nice!

api/src/mappers.js
katherine-sussol: Oh cool that's a lot easier to read :)

katherine-sussol: That's a good question! I had a look and both the Fastify request object and the raw request object don't seem to have it conveniently available without the query param. The hostname portion is, but that's not so useful for us in this instance 😿

PR #28, opened 2020-07-05
#27 add additional api test cases

### LINKED ISSUE #27: Add additional API test cases
Add test cases for mock Dgraph data. 

Now that a basic Dgraph schema has been done, would be good to add some more explicit test cases which cover all the `v1` functionality. 

To decouple the API from the actual Dgraph server (so the server doesn't have to be running to run the tests), would be good to add some nocks to intercept the calls to the Dgraph server and return the appropriate response. See https://github.com/nock/nock for more info.

wlthomson:
Fixes #27,

Completes the testing template for the v1 API, including:

- Fixtures:
  - Test data.
  - Expected graph responses.
  - Expected API responses.
- Additional `/items` tests, including examples of:
    - Using fixtures for test data, expected responses etc.
    - Simulating Dgraph responses with basic and bit more advanced `nock` functionality.

### DIFF
--- a/api/src/handlers.js
+++ b/api/src/handlers.js
@@ -25,11 +25,10 @@ const itemsHandler = (request, reply) => {
     )}'`;
     reply.code(statusCode).send(body);
   } else {
-    const dgraphClient = new DgraphClient(
-      new DgraphClientStub('http://localhost:8080')
-    );
+    const dgraphClient = new DgraphClient(new DgraphClientStub('http://localhost:8080'));
     const txn = dgraphClient.newTxn();
-    txn.query(mapRequest(request)).then(response => {
+    const { query, vars } = mapRequest(request);
+    txn.queryWithVars(query, vars).then((response) => {
       reply.code(HTTP_OK).send(mapResponse(response));
     });
   }
--- a/api/src/mappers.js
+++ b/api/src/mappers.js
@@ -40,40 +40,11 @@ export const mapRequest = (request) => {
   const { query } = request;
   const { code, name } = query;
   const exact = parseBool(query.exact);
-
-  if (code) {
-    return `{
-      query(func: has(code)) @filter(eq(code, ${code}))
-      {	
-        code
-        description
-      }
-    }`;
-  } else if (name) {
-    if (exact) {
-      return `{
-        query(func: has(code)) @filter(eq(description, ${name}) AND eq(type, "unit_of_use")) {
-          code
-          description
-        }
-      }`;
-    } else {
-      return `{
-        query(func: has(code)) @filter(regexp(description, "^${name}.*$", "i") AND eq(type, "unit_of_use")) {
-          code
-          description
-        }
-      }`;
-    }
-  } else {
-    // Default to all unit of use entities
-    return `{
-      query(func: has(code)) @filter(eq(type, "unit_of_use")) {
-        code
-        description
-      }
-    }`;
-  }
+  if (code) return { query: itemQueries.get, vars: { $code: code } };
+  else if (name) {
+    if (exact) return { query: itemQueries.searchExact, vars: { $name: name } };
+    else return { query: itemQueries.search, vars: { $name: name } };
+  } else return { query: itemQueries.all, vars: {} };
 };
 
 /**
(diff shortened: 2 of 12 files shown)


api/src/queries.js
```diff
@@ -0,0 +1,25 @@
+const items = {
+  get: `{
+        query(func: has(code)) @filter(eq(code, $code)) {	
+          code
+          description
+        }
+      }
```
katherine-sussol, line 7: I think there's a minor '}' surplus (of 1) here 🙀 . Which makes me also think it may be a good idea to add some mapper tests (though not necessarily as part of this PR)

api/src/mappers.js
```diff
@@ -32,31 +34,9 @@ export const parseRequest = (request) => {
 export const mapRequest = (request) => {
   const { query } = request;
   const { code, search } = query;
-
```
katherine-sussol, line 35: Nice, liking the separation of concerns.

api/src/handlers.js
```diff
@@ -25,11 +25,10 @@ const itemsHandler = (request, reply) => {
     )}'`;
     reply.code(statusCode).send(body);
   } else {
-    const dgraphClient = new DgraphClient(
-      new DgraphClientStub('http://localhost:8080')
-    );
+    const dgraphClient = new DgraphClient(new DgraphClientStub('http://localhost:8080'));
     const txn = dgraphClient.newTxn();
-    txn.query(mapRequest(request)).then(response => {
+    const { query, vars } = mapRequest(request);
```
katherine-sussol, line 30: Like the swap to querying with vars, unfortunately it looks like Dgraph is currently not happy with the queries with vars, I think the keys here need to be prepended with a '$' symbol? 🤑

api/src/queries.js
wlthomson: Maybe I'm going crazy, but looks like the right amount to me (just perhaps badly indented)?

wlthomson: Oops, GitHub didn't show `line 8`, you're right :).

katherine-sussol: Haha my bad, I should probably have added the comment to the last bracket, I didn't realise GitHub would just go and hide everything below the line I selected 😆

api/src/handlers.js
wlthomson: Ah yup, good catch!

api/src/queries.js
wlthomson: e43743fbd56cce4b2c628e7cab40e05ce4b5c31f.

api/src/handlers.js
wlthomson: af7ed303b302ce28dd5be40216d92d0ffc653a64.

api/src/queries.js
```diff
@@ -0,0 +1,24 @@
+const items = {
+  get: `{
```
katherine-sussol, line 2: Did some further testing and just realised we also need to declare our vars in the query! 
```suggestion
  get: `query vars($code: string){
```

api/src/queries.js
```diff
@@ -0,0 +1,24 @@
...
+    query(func: has(code)) @filter(eq(code, $code)) {	
+      code
+      description
+    }
+  }`,
+  search: `{
```
katherine-sussol, line 8: Same sort of thing as above
```suggestion
  search: `query vars($search: string) {
```

api/src/queries.js
```diff
@@ -0,0 +1,24 @@
...
+    query(func: has(code)) @filter(allofterms(description, "$search")) {
```
katherine-sussol, line 9: Some extra quotes here (using a hardcoded string needs quotes but vars don't, apparently)
```suggestion
    query(func: has(code)) @filter(allofterms(description, $search)) {
```

wlthomson:
@katherine-sussol @mark-prins 

As a follow-up to our meeting today, this PR can be considered a WIP. Will use this branch to continue to add test cases for all functionality used by mSupply.

wlthomson:
Update:

Tests added to cover functionality used by mSupply. 

Will make a new issue to duplicate tests without nocks. That should cover all of @katherine-sussol feedback, also. 

Will need to update the queries to use the recently added graph schema (also best for a standalone issue).

PR #39, opened 2020-07-08
#36 update api msupply compatibility

### LINKED ISSUE #36: Update v1 API compatibility with mSupply
Update API mappers for compatibility with mSupply usage.

mSupply only requires a subset of documented API functionality, i.e. from `universalcodes_search_by_name`:

```
// Build query for searching the name with same prefix
$t_query:="items?name="+Lowercase(String($t_search_name))+"&exact=false"
```

wlthomson:
Fixes #36.

## Description

Updates query parameters and mappers to comply with API expected by existing mSupply instances.

### Documentation:

- [x] Requires a change to the documentation.
- [ ] Documentation has been updated.

### Tests:

- [x] Requires added tests to cover changes.
- [ ] Tests have been added.
- [ ] All new and existing tests passed.

Note: tests to be added as part of existing WIP test PR (https://github.com/openmsupply/unified-codes/pull/28).

### DIFF
--- a/api/src/mappers.js
+++ b/api/src/mappers.js
@@ -2,10 +2,10 @@ const PARAMETERS = {
   '/items': ['code', 'name', 'exact'],
 };
 
-const parseBool = val => {
-  const truthyValues = [ 'true', true ];
+const parseBool = (val) => {
+  const truthyValues = ['true', true];
   return truthyValues.includes(val);
-}
+};
 
 /**
  * Parse query parameters.
@@ -66,7 +66,7 @@ export const mapRequest = (request) => {
   } else {
     // Default to all unit of use entities
     return `{
-      query(func: has(code)) @filter(eq(type,"unit_of_use")) {
+      query(func: has(code)) @filter(eq(type, "unit_of_use")) {
         code
         description
       }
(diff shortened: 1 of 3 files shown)


wlthomson:
Tests to be added as part of existing WIP test PR (https://github.com/openmsupply/unified-codes/pull/28).

wlthomson:
> Nice! It's quite interesting that mSupply uses 'name' whereas the ui uses 'search'  . I also noticed that mSupply seems to not handle whitespace (e.g. searching for "Amoxicillin oral") properly for these requests (though the api works perfectly fine - so don't think it's anything related to the PR).

Yeah, plan is to address the minor fixes in another PR. Whitespace bug is known, is due to the regex being over-simplified (in this case, because it only matches `.*`, which doesn't include any whitespace characters).

katherine-sussol:
Ah sorry - I probably didn't explain the behaviour I was seeing very clearly
In mSupply when there is a space in the param, it's not URL encoding it, e.g:
- This works:
  Screen Shot 2020-07-09 at 10 02 44 AM
- This doesn't (HTTP 400 Bad Request Response): 
  Screen Shot 2020-07-09 at 10 03 24 AM

Maybe as an aside - I'd be interested in knowing how to track pre-existing things we find from a process perspective! (e.g. if I was to create an issue in the mSupply repository would it get lost in a sea of issues?)

wlthomson:
@katherine-sussol 

Nah, you explained right, we're talking about the same thing! I thought it was due to the regex not matching correctly, but you're right that it seems to actually be due to mSupply not URL encoding special characters. Theory that this is an mSupply problem is supported by the API working as expected via the browser:


#36-update-api-msupply-compatibility-2

Will make a desktop issue for this, but a bit outside our current scope of compatibility with previous versions of mSupply, I think!

PR #46, opened 2020-07-09
#31 migrate existing data

### LINKED ISSUE #31: Migrate existing universal codes data to Dgraph
As part of the initial MVP (#2), the v1 API needs to fully comply with the API used by existing mSupply instances. As part of this, all existing data from the current implementation needs to be migration to the new Dgraph backend.

mark-prins:
Fixes #31 .

## Description
Migration code to take existing items dump from mySql and convert to dgraph compatible format.

The `sample.json` is a handy subset of data, good for debugging.

Next step is to clean the data, as advised by Emily, and to remove the orphan nodes created in this version of the script.

### Documentation:

- [x] Documentation OK.

### Tests:

- [ ] Import the file `data.v2.json` into a local dgraph and query. ( instructions in readme )
- [ ] Optionally, run the parser and generate your own version

### Query
After import, I viewed items using:

```
{
    products(func: eq(type, "medicinal_product")) @recurse(depth: 9)  {
        code
    	description
        type
        value
        has_property 
	has_child 
    }
}
```


migration/yarn.lock
```diff
@@ -0,0 +1,8 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
wlthomson, line 1: Probably best to keep consistency with `npm` throughout the repo?

migration/README.md
```diff
@@ -0,0 +1,19 @@
+# unified-codes : data migration
```
wlthomson, line 1: I think it might be better to make a separate private repo for the migration? Might be best to not make the data public?

.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
+{
```
wlthomson, line 1: Personally, I think good to keep the dev environment separate from the codebase. Need to do a tidy up, so thinking to add `.vscode` files to `.gitignore` so each user can retain their own editor configs. Thoughts?

katherine-sussol: I think there's a benefit by having some settings shared, to have some baseline consistency for how we run things (still with the flexibility to individually customise locally). Don't feel too strongly about this particular setting - though can see it helping someone later down the line who loads up Unified Codes without much context?

wlthomson: Just for context, this is a config file for vscode. Generally editor configs are included in `.gitignore` because it's a bit overkill. IMO the app shouldn't depend on any particular editor, so I think better to put scripts etc. into the `package.json`!

migration/README.md
```diff
@@ -0,0 +1,19 @@
...
+e.g. `node migrate.js ./data/items.json ./data/data.v2.json`
+
+
+### Add/update data
+
+You can use the project in the `dgraph` folder to clean and import the results
+```
+../dgraph/clean.sh
+../dgraph/alter.sh ../dgraph/sample/schema.gql 
+../dgraph/mutate.sh ./data/data.v2.json
```
katherine-sussol, line 18: I'm sorry - I think I may have deleted mutate.sh with the Dgraph GraphQL changes! (because adding data needs to be done through GraphQL layer for it to be accessible via that layer - and mutating and querying use the same endpoint in Dgraph GraphQL). Feel free to add it back in until we have a better solution (seeing all these JSON files is making it seem more worthwhile to write a script to copy JSON body into GraphQL mutation request).

.vscode/launch.json
katherine-sussol: Yeah that's fair enough, I can see the argument on that side too - particularly if open sourcing it. I'm also sensing tool/environment consistency/variance is another whole discussion point I probably don't want to get involved in just yet so happy to go with whatever works best for now :)

migration/yarn.lock
mark-prins: oh, yeah. I like yarn, that's my default. will undo.

migration/README.md
mark-prins: yeah. will remove this folder - speaks to above points about not having data in repo as well, though the migration code does need to be shared somewhere, and I think versioning is helpful for any code files.

.vscode/launch.json
mark-prins: just trying to be helpful. if anyone tries to work on the repo, then having a working debug is useful.
we'll need to discuss another way to share that info. perhaps a section in README for dev-type repos?

migration/README.md
mark-prins: all good - I'll update after moving the repo :)

.vscode/launch.json
wlthomson: I agree 100%, didn't want to come across as undermining that contribution. Would definitely agree on updating the `README.md` (or perhaps adding a wiki). @mark-prins @katherine-sussol Which one do you guys think?

PR #49, opened 2020-07-12
#32 convert api typescript

### LINKED ISSUE #32: Convert existing v1 API to typescript
Some bugs have already come up with the API, most of which would have been caught by the typescript compiler (e.g. #22). Migrating the existing code to typescript isn't a big undertaking, and I think it's a worth while trade-off in terms of time saved tracking down bugs over the lifetime of the app (also a good chance to expose some of the team to typescript). 

@mark-prins @katherine-sussol Thoughts?

wlthomson:
Fixes #32.

Last one in for week 1 sprint!

## Description

Add typescript dependencies + types to existing API code.

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.


api/src/handlers.ts
```diff
@@ -0,0 +1,61 @@
+import { DgraphClient, DgraphClientStub, Txn } from 'dgraph-js-http';
```
mark-prins, line 1: interesting that github picked up the rename of api.js but not handlers.js

api/src/mappers.ts
```diff
@@ -0,0 +1,71 @@
...
+ *     code: 3692b4bf,
+ *     name: Amoxicillin (trihydrate) 500mg solid oral dosage form"
+ *   },
+ *   {
+ *     code: 367734bf,
+ *     name: Amoxicillin & clavulanic acid 125mg &31.25mg /5mL oral liquid"
+ *   }
+ * ]
+ */
+export const mapResponse = (response: DgraphResponse): string => {
```
mark-prins, line 67: for me, tsc is complaining about the untyped objects and arrays in this method and it's also unhappy about `??`

mark-prins: I had to remove the defaults, which I think aren't necessary anyway..

```suggestion
export const mapResponse = (response: DgraphResponse): string => {
  const { data }: { data?: { query: [] } } = response;
  const { query }: { query?: [] } = data;
  return JSON.stringify(query.map(({ code, description }) => ({ code, name: description })));
};```

wlthomson: Hm, not sure why you're having a problem with these. The actual transpilation is done via babel (`tsc` just for typechecking`). What happens when you run `npm run check-types` and `npm run build`?

katherine-sussol: Interesting - `npm run check-types` and `npm run build` are both working fine for me and tests are passing on my end. `npm run debug` isn't (but that's quite easy to see why - is it worth creating a separate issue for that?). I'll have to do more reading to make sure I understand the ts stuff properly but from an execution perspective things seem to be working on my end :)

mark-prins:
ah. the penny drops!

I was testing in the terminal in vscode, and, despite setting the default for nvm to be 'lastest', my version of node was still 10.something. Running with node v13 is much better

wlthomson:
> I was testing in the terminal in vscode, and, despite setting the default for nvm to be 'lastest', my version of node was still 10.something. Running with node v13 is much better

kiddingme

Thanks for double checking!

Issue #41, opened 2020-07-08
Add external product lookup capability to GraphQL API


katherine-sussol:
## Expected Behavior

- When a GraphQL API request is made for an existing product, the GraphQL API should have the ability to return data from external sources, e.g
  - Drug Interactions (from https://rxnav.nlm.nih.gov/REST/interaction/interaction.json?rxcui=[rxcuid])
  - Drugbank details/image (from https://www.drugbank.ca/drugs/[DBID])

## Current Behavior

Not implemented yet.

## Possible Solution

- Add new query for interactions by universal code
  - Add custom resolvers in Apollo GraphQL Layer to first perform internal lookup for rxcui code, then perform external lookup against rxnav
    - (Possible enhancement) Search up/down Dgraph product hierarchy for closest matching RxCui 
- Update GraphQL schema to return external codes/urls as part of responses when returning medicinal product info

Sample drug interaction response:
- Sample response
[sampleInteractionResponse.json.zip](https://github.com/openmsupply/unified-codes/files/4935122/sampleInteractionResponse.json.zip)

Update: After discussion with Craig - we will return the whole RxNav payload as a string and let mSupply handle the display (idea being that it will be stored in mSupply and synced out to all sites). No requirement to show interactions in our Web UI as it's main benefit is the mSupply use.

wlthomson:
@katherine-sussol I think this should probably be left until next weeks sprint?

katherine-sussol:
Have some spare capacity so happy to get started on it this week - we can roll it over to target end of next sprint if we find that by end of tomorrow that it's not yet done (which is most likely what will happen).

katherine-sussol:
@wlthomson @mark-prins Wanting to get your guy's ideas on the possible solution for this one (we can also discuss it next week while planning the next sprint if this works better). It seems a lot of Dgraph's GraphQL functionality is still under development, e.g. things like custom resolvers and auth directives are not officially supported until 20.07.0. How do we feel about using features still in beta? Were there other GraphQL provider we had in mind as alternatives?

I'm also not a fan of routing GraphQL requests through our REST api since right now it's more of a necessary thing in place for backward compatibility that we eventually want to get rid of, but I think we could make it work that way if needed.

katherine-sussol:
Currently Blocked by #52

PR #55, opened 2020-07-14
#52 add apollo backend

### LINKED ISSUE #52: Add Apollo backend
While working on #41 we discovered that Dgraph GraphQL did not yet officially support custom resolvers/custom queries/auth directives. Therefore we will implement another layer over Dgraph (Apollo) to provide this support. 

(Once Dgraph officially releases we could technically revert to using Dgraph GraphQL but this may constitute unnecessary extra effort at this point).

## Expected Behaviour

- Ability to add custom GraphQL queries
- Ability to add custom resolvers (for external api lookup as part of a query)
- Auth support for specific queries/mutations (most importantly public should not be able to fiddle with our data/schema)
- [Likely deployment only concern] Dgraph should no longer be public facing once Apollo is in place (i.e. only our internal applications should be able to access it directly - otherwise this will invalidate any auth we implement in Apollo layer).

## Current Behavior

Not yet implemented.

## Possible Solution

- Add separate Apollo server to implementation as our external GraphQL api. Internally this will map to either Dgraph GraphQL+- or Dgraph GraphQL and send it off to our Dgraph instance.

katherine-sussol:
Fixes #52 

## Description

- Initial implementation of apollo backend with one query example provided (query by name - as per rest api request)
- Minimal setup - e.g. typescript, babel-watch, jest/nock/tests, commit-hooks in existing rest api haven't been added.  
  - If the general GraphQL api approach here is ok and we think it best to align with rest api, these can be done in separate PR (or happy to try to add it to this one but thought it best not to make it too huge)
- Starts up apollo-fastify-server on localhost:4000. To test:
  - Point Postman (or other GraphQL client such as Insomnia) to localhost:4000/graphql and send a request:
`{
  searchByName (text: "Amox") {
    code,
    description
  }
}`

Next Steps:
- Remove references to Dgraph GraphQL from code and docs
- Align project setup/dependencies with existing REST Api project
- Add further query support as required
- Add tests
- #41

### Documentation:

- [ ] Documentation has been added to cover changes introduced by this PR.
- [ ] This PR requires no changes to documentation.
- [ ] Documentation OK.

Probably none of the above - tidy up needed to remove references to Dgraph GraphQL both from the codebase and from the docs.

### Tests:

- [ ]  Tests have been added to cover changes introduced by this PR.
- [ ]  This PR requires no new tests.
- [ ] Tests OK.

None of the above - testing framework (should we also use jest here?) not yet added


dgraph/graphql/src/api.js
```diff
@@ -0,0 +1,34 @@
+import { ApolloServer } from 'apollo-server-fastify';
+import fastify from 'fastify';
+import { typeDefs } from './schema';
+import { dgraph } from './data';
+
+const resolvers = {
+  Query: {
+    searchByName: async (_source, _args, { dataSources }) => {
+      const dgraphQuery = `{
+        query(func: has(code)) @filter(regexp(description, /^${_args.text}.*$/i) AND eq(type, "unit_of_use")) {
```
mark-prins, line 10: :nit for the regex; it can simplify a little to `/^${_args.text}/i`
Also - the existing query only searches from the start? Pretty sure that's what we'd want anyway, just asking.

mark-prins, line 10: I have no idea why.. but for me, I had to rearrange the order before I got results:

```suggestion
        query(func: regexp(description, /^${_args.text}/i)) @filter(has(code) AND eq(type, "unit_of_use")) {
```

dgraph/graphql/src/data.js
```diff
@@ -0,0 +1,18 @@
+import { RESTDataSource } from 'apollo-datasource-rest';
+
+export class dgraph extends RESTDataSource {
+  constructor() {
+    super();
+    this.baseURL = 'http://localhost:8080';
```
mark-prins, line 6: at some other stage.. can we lift this string into a `.env` or CONSTANTS file or somewhere please?

dgraph/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "@babel/core": "^7.10.4",
+    "@babel/plugin-transform-runtime": "^7.10.4",
+    "@babel/preset-env": "^7.10.4"
+  },
+  "scripts": {
+    "build": "babel graphql/src -d graphql/lib",
+    "start": "node graphql/lib/server.js",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
```
mark-prins, line 23: :nit *mSupply Foundation*?

dgraph/graphql/src/api.js
katherine-sussol: So strange! This rearranged one still works for me so I'm happy to go with this one :) I did get a warning about needing the trigram index on description for it to work so will update the schema to reflect this.

dgraph/package.json
katherine-sussol: Ah, yes good idea - 5acc7dec8696062f2d81aad15fb9627d5128a1bb

dgraph/graphql/src/data.js
katherine-sussol: Yeah, good idea - we have a bunch that will need to eventually be moved out across both of our apis, will create a tidy up issue to track this.

dgraph/graphql/src/api.js
katherine-sussol: Thanks for the suggestion r.e. regex. Yeah, this particular query is identical/equivalent to the V1 REST API query which searches from the start (with the current mSupply behaviour I guess it makes sense). 

Thinking with the newer UI we're going to need a few new queries on the GraphQL layer (where possibly searching from the start only may not make as much sense).

mark-prins:
Looks good! ( i was about to approve again, but I think once is probably enough! )

dgraph/graphql/src/data.js
```diff
@@ -0,0 +1,18 @@
+import { RESTDataSource } from 'apollo-datasource-rest';
+
+export class dgraph extends RESTDataSource {
```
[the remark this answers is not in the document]
katherine-sussol: Yep I agree! I just noticed that a bit late actually and got merge happy, I'm adding RxNav as a datasource so will update it as part of those changes

.gitignore
```diff
@@ -1,2 +1,7 @@
+# npm
+node_modules
```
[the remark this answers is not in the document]
katherine-sussol: Ah! Sorry I'm not sure I 100% understand - do you mean changing this line to more specifically target dgraph/node_modules rather than just node_modules? Or do you mean removing the dgraph/graphql prefix from line 5?

dgraph/graphql/src/data.js
```diff
@@ -0,0 +1,18 @@
...
+  }
+
+  willSendRequest(request) {
+    request.headers.set('Content-Type', 'application/graphql+-');
+  }
+
+  async postQuery(dgraphQuery) {
+    const path = 'query';
+
+    return this.post(path, dgraphQuery);
```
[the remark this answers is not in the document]
katherine-sussol: Thanks for the feedback! Yeah, I can see the future benefit of having some of these in the constructor, particularly as we will be adding requests to /mutate later. 
Will apply in the next PR that touches this area (accessing RxNav)

.gitignore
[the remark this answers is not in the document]
katherine-sussol: Ohhh never mind I see what you meant! I just realised this .gitignore was one level above the Dgraph folder.

Issue #29, opened 2020-07-06
Add initial backend GraphQL schema


katherine-sussol:
Based on initial Dgraph schema, add scripts/schema to upload GraphQL schema to Dgraph instance.

wlthomson:
@katherine-sussol As a follow-up from our meeting. This might require some dockerisation, but I think we can handle that in a separate issue 👍.

PR #60, opened 2020-07-15
Revert "#29 add initial graph ql schema"

katherine-sussol:
Reverts openmsupply/unified-codes#33

I'd like to revert these changes, as having Dgraph GraphQL between Dgraph and Apollo GraphQL just confuses things with an extra step (and is also still in development) - so feel like it's adding an extra potential layer for things to break.


wlthomson:
Is that someone trying to do a REVERT!?

gotya

Agree with the idea, but wondering if might be better to make an issue and do a PR to remove the stuff no longer needed. Only reason I say this is:

1. My own experiences with reverts is they often mess up stuff by making changes you didn't anticipate!
2. For issue tracking etc., it's useful to have an issue + PR for seeing whats been changed and why.

Shouldn't be too much work hopefully, can just keep this PR open and just go through the change log and delete what needs to be removed? If I'm wrong and your PR change log ends up identical to this, then you win the prize!

moneycat

katherine-sussol:
Very appropriate cat gifs :). I'll give a manual PR a go, I was being hopeful it could just be done in one revert, but do see one thing that needs to be preserved that's currently being undone (the extra codes for amox tablets and paracetamol tablets).

katherine-sussol:
Closed as merged in via separate PR #62

PR #63, opened 2020-07-15
Add checkboxes to PR Template

katherine-sussol:
The instructions to tick something when there was only one option was bugging me when creating my last PR 🙀  - did I get the original intention correct here?

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -16,16 +16,14 @@ Fixes #.
 
 Tick below if one of the following applies:
 
-1. Documentation has been added to cover changes introduced by this PR.
-2. This PR requires no changes to documentation.
-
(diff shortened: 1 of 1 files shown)


wlthomson:
Hmm, the idea was to only have one checkbox which if ticked means the task is done (i.e. documentation for this PR is done). Having two checkboxes makes it hard to tell if the PR has completed the tasks or not (that's why I reduced it down to one checkbox). 

I think good rule of thumb is to make an issue for all PRs, don't like advocating to close PRs when people have worked on them :(.

katherine-sussol:
Fair enough, this one was a really small thing so wasn't sure if we wanted an issue for it. I don't mind closing it - at least on the same page R.E the template now!

wlthomson:
> Fair enough, this one was a really small thing so wasn't sure if we wanted an issue for it. I don't mind closing it - at least on the same page R.E the template now!

smilecat

PR #72, opened 2020-07-19
#41 add drug interaction lookup

### LINKED ISSUE #41: Add external product lookup capability to GraphQL API

katherine-sussol:
Fixes #41 

## Description

- Applied following feedback from prior PR  
  - Git Ignore: https://github.com/openmsupply/unified-codes/pull/55#discussion_r454751477
  - Class Naming: https://github.com/openmsupply/unified-codes/pull/55#discussion_r454764168
  - Class Constructor: https://github.com/openmsupply/unified-codes/pull/55#discussion_r454768144
-  New query added to schema (getDrugInteractions with universal code as input parameter)
- Custom resolver added to first query Dgraph for the RXCui, then query RxNav

**Future Improvements**
- Had some discussion with Craig and this sort of aligns with the previous demo on interactions, although it is now a GraphQL request/response rather than REST. Future improvement may be to structure this better (return properties rather than huge JSON string) and implement a backward compatible version that returns the huge JSON string in the REST API.
- Possibly we may want to expand the functionality to search parent/child nodes for a RXCui in the DB, currently we'll only return results if the currently queried node has RXCui as a direct property.

Sample Custom Query
`{
  getDrugInteractions (code: "368C74BF")
}`

Sample Response
`{
"data": { 
"getDrugInteractions": { [RxNav JSON Payload]
}
}
}`

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.

### DIFF
--- a/dgraph/graphql/src/api.js
+++ b/dgraph/graphql/src/api.js
@@ -9,8 +9,8 @@ const server = new ApolloServer({
   resolvers,
   dataSources: () => {
     return {
-      dgraphDataSource: new DgraphDataSource(),
-      rxNavDataSource: new RxNavDataSource()
+      dgraph: new DgraphDataSource(),
+      rxNav: new RxNavDataSource()
     };
   },
 });
--- a/dgraph/graphql/src/api.js
+++ b/dgraph/graphql/src/api.js
@@ -1,7 +1,7 @@
 import { ApolloServer } from 'apollo-server-fastify';
 import fastify from 'fastify';
 import { typeDefs } from './schema';
-import { dgraph } from './data';
+import { DgraphDataSource } from './data';
 
 const resolvers = {
   Query: {
@@ -23,7 +23,7 @@ const server = new ApolloServer({
   resolvers,
   dataSources: () => {
     return {
-      dgraphDataSource: new dgraph(),
+      dgraphDataSource: new DgraphDataSource(),
     };
   },
 });
--- a/dgraph/graphql/src/api.js
+++ b/dgraph/graphql/src/api.js
@@ -1,22 +1,8 @@
 import { ApolloServer } from 'apollo-server-fastify';
 import fastify from 'fastify';
 import { typeDefs } from './schema';
-import { DgraphDataSource } from './data';
-
-const resolvers = {
-  Query: {
-    searchByName: async (_source, _args, { dataSources }) => {
-      const dgraphQuery = `{
-        query(func: regexp(description, /^${_args.text}/i)) @filter(has(code) AND eq(type, "unit_of_use")) {
-          code
-          description
-        }
-      }`;
-      const resp = await dataSources.dgraphDataSource.postQuery(dgraphQuery);
-      return resp.data.query;
-    },
-  },
-};
+import { DgraphDataSource, RxNavDataSource } from './data';
+import { resolvers } from './resolvers'
 
 const server = new ApolloServer({
   typeDefs,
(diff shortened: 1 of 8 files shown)


dgraph/graphql/src/api.js
```diff
@@ -1,29 +1,16 @@
...
+import { DgraphDataSource, RxNavDataSource } from './data';
+import { resolvers } from './resolvers'
 
 const server = new ApolloServer({
   typeDefs,
   resolvers,
   dataSources: () => {
     return {
-      dgraphDataSource: new dgraph(),
+      dgraphDataSource: new DgraphDataSource(),
```
wlthomson, line 12: Can probably just call this `dgraph`, `rxNav`?

dgraph/graphql/src/schema.js
```diff
@@ -1,6 +1,8 @@
 import { gql } from 'apollo-server-fastify';
 
 const typeDefs = gql`
+  scalar JSON
```
wlthomson, line 4: 😨.

katherine-sussol: Haha, yeah I had an initial structure in mind and had mapped them in a spreadsheet, but it was sounding like we wanted all the data to be stored in mSupply and would then be synced out to all sites using mSupply's syncing). I think we may need a wider discussion on it.

This is what I originally had for the GraphQL Schema:
`  type DrugInteraction {
    name: String
    description: String
    severity: String }`
Possibly including the interacting drug's rxcui here as well

dgraph/graphql/src/api.js
katherine-sussol: Hmm good point, I think we can call the variables here just dgraph and rxNav for brevity.
Updated with https://github.com/openmsupply/unified-codes/pull/72/commits/525700ac54637cac82059a19fd5760a92b41e00e

dgraph/graphql/src/schema.js
wlthomson: > ```
> type DrugInteraction {
>     name: String
>     description: String
>     severity: String
>     source: String
>  }
> ```
> 
> Possibly including the interacting drug's rxcui here as well

Yeah, this is more like what I had in mind. I don't think we need to worry too much about what mSupply does with the data when it gets it (at least at this stage). Since mSupply will be getting the data via the legacy API, I think a good idea is to do they JSON conversion there (as part of this will need to update the legacy API to query GraphQL endpoints instead of Dgraph).

PR #75, opened 2020-07-21
#54 add UI table component

### LINKED ISSUE #54: Add UI table component
Add table component for displaying drug data.

## Expected Behavior

E.g. https://www.drugbank.ca/drugs.

## Current Behavior

N/A.

## Possible Solution

For now, no styling (will need to make some decisions about this later, for now can just stick to standard defaults).

wlthomson:
Fixes #54.

## Description

Adds components + stories for basic entity table.

Note: for background on component hierarchy, see https://bradfrost.com/blog/post/atomic-web-design/. Base components are redefined to make it clear which atomic components are used by the application (a bit overkill atm, but will scale out nicely, I think).

### Documentation:

- [ ] Documentation OK.

### Tests:

- [x] Tests OK.


web/stories/EntityTableRow.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import { EntityTableRow } from '../src/components';
+
+const entities = {
```
mark-prins, line 4: 3rd time these have shown up! time to move to a separate file, I think.

web/src/components/organisms/entity/EntityTable.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const EntityTable: EntityTable = ({ tableProps, headProps, headerProps, bodyProps, rowProps, entities }: { tableProps: TableProps, headProps: TableHeadProps, headerProps: EntityTableHeaderProps, bodyProps: TableBodyProps, rowProps: EntityTableRowProps, entities: Entity[] }) => {
+    const mapEntity = (entity: Entity) => <EntityTableRow {...{...rowProps, entity}}></EntityTableRow>;
+    
+    const EntityTableRows = React.useCallback(() => (
+        <React.Fragment>{entities.map(mapEntity)}</React.Fragment>
+    ), [rowProps, entities]);
+    
+    return (
+        <Table {...tableProps}>
+            <TableHead {...headProps}>
```
mark-prins, line 26: I think neater to have the `TableHead` inside the `EntityTableHeader` as the latter only makes sense in the context of a table head and it tidies up this component which is getting unwieldy.
In which case the complimentary `EntityTableBody` would make sense - a `TableBody` which includes the `EntityTableRows`

web/src/components/atoms/core/Table.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
```
mark-prins, line 1: Are these in for future use?

web/src/components/organisms/entity/EntityTable.tsx
wlthomson: That's actually initially how I had it, but I ended up changing it because it broke consistency with the general table API (e.g. `TableBody` should be a context, not include rows). 

As these are `organisms`, they could be considered at a higher level of abtraction but I think they need a name that makes it clearer that they are compositions of  `Head`+`Header` and `Body`+`Row[]`. Can go with `EntityTableHeader = TableHead + EntityTableRow` as that's unambiguous, but need an equivalent for `EntityTableBody` (something like `EntityTableData", but I don't like that very much!) Any ideas?

web/src/components/atoms/core/Table.tsx
wlthomson: Yeah, see description.

> Base components are redefined to make it clear which atomic components are used by the application (a bit overkill atm, but will scale out nicely, I think).

By redefining all the core components we use, it gives us a nice clear view of all the atoms our app is built on. These will change a bit (to include app styles etc.), so it's also useful for future usage.

web/stories/EntityTableRow.stories.tsx
wlthomson: Yeah. purposefully kept these declarations WET as they are duplicates of the test data for the legacy API (and the Dgraph backend data). I think leave the DRY implementation for `nx`?

katherine-sussol:
Ah! Nice, thanks for setting up and adding me as a reviewer. I don't really have any useful feedback at this stage, but managed to get it running and have a browse around. Gives me a good idea of the general approach for building out components :).

PR #78, opened 2020-07-22
#67 update graph ql schema

### LINKED ISSUE #67: Update GraphQL schema with entity query
Currently the Apollo GraphQL layer only exposes a very simple query mimicking the REST API (returning only matching products with their code and description).

We'll need to expand this to return the whole underlying structure. Ideally we will return everything in the DB and the client side can choose what filtering to apply.

## Expected Behavior

Ability to see properties and children of products in the results returned by the GraphQL layer

## Current Behavior

N/A 

## Possible Solution

Add resolver for Entity type that returns all entities (in tree structure) and properties.
Possibly will need to add custom types for external properties and then move custom resolver logic into the resolver for that custom type.

katherine-sussol:
Fixes #67 

## Description

- Replaced search by name query with query for all entities
- Added optional search filter (for type/code) - type will be necessary to retrieve non-medicinal products unless we add a parent 'products' node to our DB (which sits above medicinal_products and non medicinal products)
- Removed drug interactions query as per discussion - this will be added to the REST API (likely a low priority for Graph API as mSupply will query through REST). 
- NB: It may be tricky to know when and when not to request properties from the client side with this structure, as it is very generic - starting to possibly see the benefit of better typing! However this is just my observation...

**Sample queries (these will now need to be recursive):** 

- Generic request for all products and some properties
```
query GetProducts {
  entities {
    code 
    description
    type
    has_child {
      code
      description
      type
      has_child {
        code
        description
        type
        has_child {
          code
          description
          type
          has_property {
            type
            value
            has_property {
              type
              value
            }
          }
        }
      }
    }
  }
}
```
- Request for products by code
```
query GetProductsByCode {
  entities (filter: { code: "GH89P98W" } ) {
    code 
    description
    type
    has_child {
      code
      description
      type
      has_child {
        code
        description
        type
        has_child {
          code
          description
          type
        }
      }
    }
  }
}
```
### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.

### DIFF
--- a/dgraph/graphql/src/dgraph/queries.js
+++ b/dgraph/graphql/src/dgraph/queries.js
@@ -13,7 +13,7 @@ export const queries = {
   },
   entities: (type) => {
     return `{
-      query(func: has(code)) @filter(eq(type, ${type})) @recurse(loop: false)  {
+      query(func: eq(type, ${type})) @filter(has(code)) @recurse(loop: false)  {
         code
         description
         type
--- a/dgraph/graphql/src/dgraph/queries.js
+++ b/dgraph/graphql/src/dgraph/queries.js
@@ -1,0 +1,19 @@
+export const queries = {
+  entities: (filter) => {
+    const type = filter?.type ?? "medicinal_product";
+    const codeFilter = filter?.code ? `@filter(eq(code, "${filter.code}"))` : ``;
+
+    return `{
+      query(func: eq(type, ${type})) ${codeFilter} @recurse(loop: false)  {
+        code
+        description
+        type
+        value
+        has_child
+        has_property
+      }
+    }`
+  }
+}
+
+export default { queries };
--- a/dgraph/graphql/src/dgraph/queries.js
+++ b/dgraph/graphql/src/dgraph/queries.js
@@ -13,7 +13,7 @@ export const queries = {
   },
   entities: (type) => {
     return `{
-      query(func: eq(type, ${type})) @filter(has(code)) @recurse(loop: false)  {
+      query(func: eq(type, ${type})) @filter(has(description)) @recurse(loop: false)  {
         code
         description
         type
--- a/dgraph/graphql/src/dgraph/queries.js
+++ b/dgraph/graphql/src/dgraph/queries.js
@@ -1,10 +1,19 @@
 export const queries = {
-  entities: (filter) => {
-    const type = filter?.type ?? "medicinal_product";
-    const codeFilter = filter?.code ? `@filter(eq(code, "${filter.code}"))` : ``;
-
+  entity: (code) => {  
+      return `{
+        query(func: eq(code, ${code})) @recurse(loop: false)  {
+          code
+          description
+          type
+          value
+          has_child
+          has_property
+        }
+      }`
+    },
+  entities: (type) => {
     return `{
-      query(func: eq(type, ${type})) ${codeFilter} @recurse(loop: false)  {
+      query(func: has(code)) @filter(eq(type, ${type})) @recurse(loop: false)  {
         code
         description
         type
--- a/dgraph/graphql/src/dgraph/queries.js
+++ b/dgraph/graphql/src/dgraph/queries.js
@@ -1,6 +1,6 @@
 export const queries = {
-  entity: (code) => {  
-      return `{
+  entity: (code) => {
+    return `{
         query(func: eq(code, ${code})) @recurse(loop: false)  {
           code
           description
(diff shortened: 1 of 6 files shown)


dgraph/graphql/src/dgraph/queries.js
```diff
@@ -0,0 +1,19 @@
+export const queries = {
+  entities: (filter) => {
+    const type = filter?.type ?? "medicinal_product";
+    const codeFilter = filter?.code ? `@filter(eq(code, "${filter.code}"))` : ``;
+
+    return `{
+      query(func: eq(type, ${type})) ${codeFilter} @recurse(loop: false)  {
```
katherine-sussol, line 7: I just realised this won't work if we want to search by code, but don't know what level in the tree we're looking at. I'm thinking the filtering we want here is dependent on what we want to do in the UI (or the UI can also request everything and apply it's own filtering...but could become quite large).

dgraph/graphql/src/dgraph/queries.js
```diff
@@ -0,0 +1,19 @@
...
+        code
+        description
+        type
+        value
+        has_child
+        has_property
+      }
+    }`
+  }
+}
```
wlthomson, line 17: I think a few things here:

1. `code` is always unique, so can be a different query.
2. Let the resolver parse the args (rather than one big `entities` method which does everything). We will need to have a think about how to handle all the different filtering possibilities without combinatoric blowout).

My suggestion:

```suggestion
export const queries = {
  entity: code => // get entity by code
  entities: type => // get entities by type
}
```

See below for resolver suggestion,

dgraph/graphql/src/resolvers.js
```diff
@@ -1,39 +1,14 @@
...
-          code
-          description
-        }
-      }`;
+    entities: async (_source, _args, { dataSources }) => {
+      const { filter } = _args;
+      const dgraphQuery = queries.entities(filter);
       const response = await dataSources.dgraph.postQuery(dgraphQuery);
       return response.data.query;
     },
```
wlthomson, line 10:
```suggestion
    entity: async (_source, _args, { dataSources }) => {
      const { code } = _args;
      if (!code) return null;
      const query = queries.entity(code);
      const response = await dataSources.dgraph.postQuery(query);
      return response.data.query;
    },
    entities: async (_source, _args, { dataSources }) => {
      const { filter } = _args;
      const { type } = filter;
      if (!type) return null;
      const query = queries.entities(type);
      const response = await dataSources.dgraph.postQuery(query);
      return response.data.query;
    }
```

dgraph/graphql/src/dgraph/queries.js
katherine-sussol: Cool, yeah I was thinking along the same lines after realising the inadvertent bug 😸 . Good point r.e. moving args into resolvers, will do.

Also realised there's a bit of a dodgy bug where the current Dgraph query lets you query for a type of 'strength' and returns properties instead of entities (which then confuses Apollo), will fix this too.

katherine-sussol: Fixes applied in https://github.com/openmsupply/unified-codes/commit/b5eebe44775bff017d8b236cbf95e78afbcee3ac

dgraph/graphql/src/resolvers.js
katherine-sussol: Thanks for the suggestion! I made code a required parameter in the schema for single entity requests (Apollo will return HTTP 400 / GraphQL validation error) as I think it doesn't make much sense to be searching for a specific entity without one. Aside from that I've pretty much implemented the above as is with https://github.com/openmsupply/unified-codes/commit/b5eebe44775bff017d8b236cbf95e78afbcee3ac

katherine-sussol:
Sorry - I need to be a bit more thorough with my testing before sending out the PRs - I get excited when I see the main use case working :) I've since fixed a few additional bugs I found. Query for code should now work like below:

**Query**
```
query GetProductsByCode {
  entity (code: "GH89P98W") {
    code 
    description
    type
    has_child {
      code
      description
      type
      has_child {
        code
        description
        type
        has_child {
          code
          description
          type
        }
      }
    }
  }
}
```

**Response**
```
{
  "data": {
    "entity": {
      "code": "GH89P98W",
      "description": "Paracetamol",
      "type": "medicinal_product",
      "has_child": [
        {
          "code": null,
          "description": "Paracetamol oral",
          "type": "form_category",
          "has_child": [
            {
              "code": "P4R40T4B",
              "description": "Paracetamol oral tablet",
              "type": "form",
              "has_child": [
                {
                  "code": "5589F4BF",
                  "description": "Paracetamol oral tablet, 300mg",
                  "type": "unit_of_use"
                }
              ]
            }
          ]
        }
      ]
    }
  }
}
```

PR #80, opened 2020-07-23
#79 update data service project structure

### LINKED ISSUE #79: Refactor dgraph project structure
Discussion issue for reorganising/refactoring the dgraph project.

## Expected Behavior

For discussion.

## Current Behavior

```
/dgraph
  /graphql
    /lib
    /src
  /sample
  foo.sh
  bar.sh
  ...
```

## Possible Solution

Current thoughts are something like:

```
/data-service
  /database
    /data
    /scripts
      foo.sh
      bar.sh
      ...
  /server
    /lib
    /src
```

wlthomson:
Fixes #79.

## Description

Reorganise `dgraph` project and rename to `data-service`.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [x] Tests OK.


wlthomson:
@katherine-sussol @mark-prins Not sure what happened there... fixed up with a rebase and somehow nothing exploded! 

Looks like some tests still failing so will investigate, probably just need to point them to the new directory.

data-service/server/src/resolvers.js
```diff
@@ -0,0 +1,46 @@
+const queries = {
```
katherine-sussol, line 1: Queries look good - I think I'm seeing double though 👓  are these the same as the ones in queries.js?

data-service/server/lib/api.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
```
katherine-sussol, line 1: I saw the lib/ directory was already updated in `.gitignore`, is it also worth removing the files from source control?

data-service/package.json
```diff
@@ -16,10 +16,8 @@
     "@babel/preset-env": "^7.10.4"
   },
   "scripts": {
-    "build": "babel graphql/src -d graphql/lib",
-    "start": "node graphql/lib/server.js",
+    "build": "babel server/src -d server/lib",
+    "start": "node server/lib/server.js",
     "test": "echo \"Error: no test specified\" && exit 1"
-  },
-  "author": "mSupply Foundation",
```
katherine-sussol, line 23: Hm (this is just a question), is this something that has a convention yet, or pending discussion on a convention? 
I took a quick look through the other projects and noticed quite a bit of a variety. E.g. mobile uses
 "Sustainable Solutions (NZ) Ltd", and V1 of UC has Chris listed as the author 😆 .

mark-prins: No convention yet ( clearly! ) 
The direction I would expect is for everything to be under `The mSupply Foundation` or the `mSupply Foundation`? So yes, still a little discussion required

wlthomson:
@katherine-sussol 

Re. `.gitignore`: `lib` files are removed in https://github.com/openmsupply/unified-codes/pull/88/commits/b999f4164c10e302840a6dc4d52fe109318d5f94 (not included in this PR).

Re. `package.json`: just updated this for consistency with other projects. Since we haven't agreed on how to do this, I think best to keep stuff consistent (can update all the projects at a later time).

katherine-sussol:
Re. `.gitignore` cool - hadn't seen those commits yet 👍
Re. `package.json` fair enough, my question was more out of self-interest next time when creating a new project (and for the nx repo update - will follow suit)
Re. `queries` all good, I'm personally not fussed over same vs separate file at this point (with just 2 queries). Would it make sense to delete the standalone queries.js in this case then, to avoid confusion?

wlthomson:
@katherine-sussol  `queries` file should be deleted (maybe in a later PR, sorry!). Might be best to revisit after all PRs are merged.

PR #89, opened 2020-07-27
#48 initial repo wiki

### LINKED ISSUE #48: Add repo wiki
Tidy up for work introduced as part of weekly sprint (#30).

## Expected Behavior

- "Helper" config files (e.g. `.vscode` files) listed in repo wiki.
- Data files not duplicated in repo (e.g. only `data.json` or `data.rdf`, not both).

## Current Behavior

- No repo wiki.
- Repo contains config files and duplicate data files.

## Possible Solution

- Remove config files and unneeded data files from repo.
- Add repo wiki.
- Add config examples to wiki.

katherine-sussol:
Fixes #48.

## Description

Minimal proposal for some initial wiki pages. Not planning on merging this PR in, but rather applying the content in the .md files to the wiki (annoyingly Github doesn't store the wiki files in the repo itself).

There are corresponding changes to remove launch.json from git but this PR is wiki only (to avoid having to try to do a partial PR merge).

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.


wiki/Home.md
```diff
@@ -0,0 +1,26 @@
+
+Welcome to the unified-codes wiki!
+
+"unified-codes" is intended to complately replace the original universal drug codes server.
```
mark-prins, line 4:
```suggestion
"unified-codes" is intended to completely replace the original universal drug codes server. 
```

wiki/Home.md
```diff
@@ -0,0 +1,26 @@
...
+<!-- This is quite high level - A diagram may be useful -->
+* Database 
+  * [Dgraph](https://dgraph.io/docs/)
+* APIs 
+  * GraphQL served by [Apollo](https://www.apollographql.com/docs)
+  * REST served by [Fastify](https://www.fastify.io/)
+* Web UI 
+  * [React](https://reactjs.org/docs)
+* Authentication
+  * TBC
```
mark-prins, line 18: Integration with an open ID compliant server, using JSON web tokens. Currently implemented using [Keycloak](https://www.keycloak.org/) as the authentication server ( see the data-service sub-folder for the 'get-you-up-and-running' docker implementation )

wiki/Home.md
```diff
@@ -0,0 +1,26 @@
+
+Welcome to the unified-codes wiki!
+
+"unified-codes" is intended to complately replace the original universal drug codes server. 
+* The "original" universal drug codes server is currently located at https://sussol.github.io/universal-drug-codes-server/.
+* "unified-codes" (aka V2) is not yet deployed.
```
wlthomson, line 6: Not sure how others feel, but I think a design principle for the wiki should be that it is as "user-friendly" as possible, e.g. something we can show to stakeholders/funders. In that respect, I think mentions of "unified-codes" versus "universal codes drug server" could cause quite a lot of confusion for users. IMO we can just differentiate on the basis of `v1` and `v2`, and would be good to keep discussion of `v1` to the legacy API documentation. 

If others agree, my suggestion would be something like this (but you can probably write it better!)

```
Welcome to the unified-codes wiki!

[Some info on unified-codes, e.g. "This wiki is intended as the main reference document/information hub for the unified-codes project"].

Note: the old `v1` universal codes server has been retired, but can still be found [here](https://sussol.github.io/universal-drug-codes-server/).
```

wiki/Home.md
```diff
@@ -0,0 +1,26 @@
...
+
+## System Overview/Design [TO COMPLETE]
+<!-- This is quite high level - A diagram may be useful -->
+* Database
```
wlthomson, line 10: I think might be better to keep the database and GraphQL layer together (since they function as one service, i.e. the data service)? I think easier to understand and more consistent with the codebase and other documentation.

wiki/Home.md
```diff
@@ -0,0 +1,26 @@
...
+
+"unified-codes" is intended to complately replace the original universal drug codes server. 
+* The "original" universal drug codes server is currently located at https://sussol.github.io/universal-drug-codes-server/.
+* "unified-codes" (aka V2) is not yet deployed.
+
+## System Overview/Design [TO COMPLETE]
+<!-- This is quite high level - A diagram may be useful -->
+* Database 
+  * [Dgraph](https://dgraph.io/docs/)
+* APIs
```
wlthomson, line 12: As above, can just go with `Legacy API` here, I think.

wiki/graphql-api.md
```diff
@@ -0,0 +1,200 @@
+## GraphQL API
+
+We expose a GraphQL API for universal drug codes using [Apollo Server](https://github.com/apollographql/apollo-server#readme).
```
wlthomson, line 3: Since this is documenting a service, I think maybe better to just focus on the API (hide implementation details like Apollo etc.)? I think the user will mostly just want to know how to get the data they want, implementation detail is probably more suitable when documenting the architecture or contributor guides etc.

wiki/Home.md
mark-prins: Sorry, I thought that we had a diagram to hand which we could drop in here. But no. It's a good idea though & we'll need to knock something up, maybe next week.

wiki/graphql-api.md
mark-prins: So.. the wiki is more to explain how to format queries and how the data is structured?
We should create an area for more under-the-hood explanations somewhere else?

wlthomson: I would still suggest having the implementation details in the wiki, just not in the top-level section, as users will just want to know how it works, rather than whats under-the-hood. I think implementation details might be nice to be left to a section for contributors/"building from source"?

wiki/Home.md
katherine-sussol: Whoops! Thanks

wiki/Home.md
katherine-sussol: Yep, agree with being "user-friendly" (though what dev consumers find friendly may be quite different to stakeholders 😆 ). I think splitting it into sections as you've suggested below will help with this distinction, and we can keep these overview type sections more focused on a non-technical audience. Also see how unified-codes vs universal codes is kind of confusing - I thought it may just help people looking for the implementation details to know it also has another name (but we can put the details in a separate section).

wiki/graphql-api.md
katherine-sussol: Cool. So this means top-level things will be end-user focused (user guides/query and response structure) and I'll move the details (including the vscode config file) into a separate section for developers/contributors?

wiki/Home.md
wlthomson: Sorry, should have clarified! When I say user-friendly, I mean for general users of the services (e.g. avoid language that won't be understood by the average user). I think a good general pattern for documentation is to keep things simple at the top-level and leave implementation/technical details for more specific sections (e.g. "contributors", "technical reference", "advanced usage" etc.) Just opinion based on my experience!

wiki/graphql-api.md
wlthomson: thumbsup

wiki/Home.md
katherine-sussol: Yeah, that makes sense :). I'm not used to putting everything in one place - my pattern of thinking was separate docs/locations for different audiences but what you've described sounds like it would work well for the wiki!

wiki/Home.md
katherine-sussol: Oh nice, thanks for the snippet. Added it to the TBC section for now while finding a home for these details

wiki/Home.md
katherine-sussol: @wlthomson Yep - makes sense if GraphQL/DB are bundled as per above
@mark-prins Yeah - I had an initial go with https://mermaid-js.github.io/mermaid/#/ as it was in the stackedit template but it was embarrassingly hard to read.

katherine-sussol:
Will close the PR for now - applying to the wiki (with feedback - thanks for the comments)

Issue #86, opened 2020-07-26
Add EntityBrowser component


wlthomson:
Add stateful component for grouping `EntityTable` and `SearchBar` components.

## Expected Behaviour

- `EntityBrowser` component, wraps `EntityTable` and `SearchBar` components for displaying and querying data.
-  `entities` prop containing entity data to be displayed/queried.

## Current Behaviour

- `EntityTable` component for displaying entities.
- `SearchBar` component for registering user input and clear/search click events.

## Possible Solution

- Wrap `EntityTable` and `SearchBar` components.
- Maintain table data state and filter based on `onChange/onClick` events.

PR #90, opened 2020-07-28
#66 backend authorisation logic

### LINKED ISSUE #66: Backend authorisation logic
Need to authenticate users against an authentication store and be able to limit actions in dgraph, such as limiting mutations to specific roles.

## Expected Behavior
Should be able to restrict access to part of the web application to authenticated users. No requirement initially to have fine grained permissions, this will come later.

mark-prins:
Fixes #66 
Branched off #86, best to wait for that to be merged before reviewing. Will make your life easier!

## Description
Accepts a JWT token in the header of a request and parses it to determine if the user is logged in.
Adds a docker image for keycloak and imports a sample configuration on running this, which allows for testing.
Creates a graphql query for `user` in order to test authentication and decoding of the JWT

### Documentation:

- [ ] Documentation OK.

### Tests:

No automated tests yet.
Here are some things to try:

#### Sign In
**Configuration**
Property | Value
:-- | :--
URL | http://localhost:9990/auth/realms/unified-codes/protocol/openid-connect/token
Method | POST

**Body**
Key | Value
:-- | :--
client_id | unified-codes-data
grant_type | password
client_secret | 55ed8757-d515-4e3e-86ac-c05fa1151393
username | admin
password | pass

**Tests**
- [ ] Able to login when providing username `admin` and password `pass`
- [ ] Fail to login if username incorrect
- [ ] Fail to login if password incorrect
- [ ] Fail to login if secret incorrect
- [ ] Returns a JWT if all is well

#### Query User
**Configuration**
Property | Value
:-- | :--
URL | http://localhost:4000/graphql
Method | POST

Take the value of the `access_token` property of the token response. Use this when submitting the next query.
You can also validate the token using `jwt.io` if you first fetch the public key. This is available on `http://127.0.0.1:9990/auth/realms/unified-codes` or via the admin interface.

Remember to sandwich the resulting public key with 
```
-----BEGIN PUBLIC KEY-----
[public key]
-----END PUBLIC KEY-----
```
when adding this to `jwt.io` when testing.

**Body**
```
{ 
    user{
        name
    }
}
```

**Tests**
- [ ] Able to retrieve user name if all is ok
- [ ] If token is expired then returns an error
- [ ] If no token provided then returns an error


mark-prins:
Note: had the app fetch the public key dynamically and then got stuck with handling this synchronously. have given up for now, and just put the key in the app. Will change that once I figure it out.

wlthomson:
Have had a look and mostly LGTM, but might wait for a few more merges, as it's a bit tricky with such a big diff!

wlthomson:
Tried to get this running this arvo, but also having a bit of trouble with config 😢. Will give it another go tomorrow morning!

data-service/authentication/docker.sh
```diff
@@ -0,0 +1 @@
+docker run -p 9990:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin -e KEYCLOAK_IMPORT=/tmp/unified-codes.json --mount type=bind,source="$(pwd)",target=/tmp jboss/keycloak
```
wlthomson, line 1: For maintainability and testing etc. thinking that instead of importing, it might be helpful to have some scripts which generate the realm config? Would also considerably reduce the number of lines being committed!

please

data-service/server/src/api.js
```diff
@@ -3,24 +3,77 @@ import fastify from 'fastify';
 import fastifyCors from 'fastify-cors';
 import { typeDefs } from './schema';
 import { DgraphDataSource, RxNavDataSource } from './data';
-import { resolvers } from './resolvers'
+import { resolvers } from './resolvers';
+import jwt from 'jsonwebtoken';
+// import fetch from 'node-fetch';
+
+const AUTH_BASE_URL = 'http://127.0.0.1:9990/auth/realms/unified-codes';
+const PUBLIC_KEY = `-----BEGIN PUBLIC KEY-----
```
wlthomson, line 11: Wonder if we could use some sort of local storage/caching for this? I think we have a registered key pair for our domain, could we use that?

data-service/server/src/api.js
```diff
@@ -3,24 +3,77 @@ import fastify from 'fastify';
...
+// ${authConfig.public_key}
+// -----END PUBLIC KEY-----`;
+// };
 
 const server = new ApolloServer({
   typeDefs,
   resolvers,
   dataSources: () => {
     return {
       dgraph: new DgraphDataSource(),
```
wlthomson, line 60: Wonder if we could make `keycloud` a data source?

data-service/server/src/api.js
```diff
@@ -3,24 +3,77 @@ import fastify from 'fastify';
...
-      rxNav: new RxNavDataSource()
+      rxNav: new RxNavDataSource(),
     };
   },
+  context: ({ req }) => {
+    const token = getToken(req.headers);
+    const user = getUser(token);
```
wlthomson, line 66: Would be quite nice here to abstract away the token, e.g.

```
const user = getUser(req)
```

data-service/server/src/resolvers.js
```diff
@@ -24,6 +26,11 @@ const queries = {
     }`;
   },
 };
+const userHasRole = (user, role) => {
+  if (!user.roles || !user.roles.length) return;
```
wlthomson, line 30: Should probably return `false` here?

wlthomson, line 30: If we have babel setup with null coarsening operator, can do `if (!user?.roles?.length) ...`

data-service/server/src/api.js
```diff
@@ -3,24 +3,77 @@ import fastify from 'fastify';
...
+      return;
+    }
+    //const publicKey = await fetchPublicKey();
+    const decodedToken = jwt.verify(token, PUBLIC_KEY, { algorithms: ['RS256'] });
+
+    // saving this for later
+    // const hasAdminRole = decodedToken.hasRealmRole(ADMIN_ROLE);
+    const { realm_access, name, given_name, family_name } = decodedToken;
+    const roles = realm_access ? realm_access.roles || [] : [];
+    const user = { name, given_name, family_name, roles };
```
wlthomson, line 27: Would be nice to have a `User` class maybe?

Can then do stuff like:

```
const user = new User(name, roles);
...
`user.hasRole(role)` etc.
```

hmm_1

data-service/authentication/docker.sh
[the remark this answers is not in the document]
mark-prins: 0060579

mark-prins: 54ab5ab
mac only at the moment - running docker in a sep terminal so that I can run it all at once.
Also the `Realm Role` assignment doesn't work. Other users report that it worked in previous versions 🤷‍♂️ 
For now - manual role assignment. The alternative is to request user details and extract the user guid so that roles can be assigned. not sure that's warranted.

data-service/server/src/api.js
mark-prins: 4792599

data-service/server/src/resolvers.js
mark-prins: 4792599

mark-prins: 4792599

data-service/server/src/api.js
mark-prins: 4792599

data-service/server/src/api.js
mark-prins: Created #96 - worried that this PR is getting too big for its boots.

data-service/server/src/api.js
mark-prins: will have a look in #96

PR #109, opened 2020-08-06
#108 product search failure with short term

### LINKED ISSUE #108: Product search fails with less than 3 chars in search term
## Expected Behaviour

User should be able to search via legacy API for a product by name with less than 3 chars in the search term.

Example:
`GET {RESTHost}/items?name=Am`
should return all results beginning with Am (e.g. Amoxicillin oral capsule, 250mg)

## Current Behaviour

Request times out, with the following error logged (originating from Dgraph):
`APIError: : regular expression is too wide-ranging and can't be executed efficiently
`

## Possible Solution

Random thoughts here, don't necessarily think any of these are ideal 😿 
- Block requests with less than 3 chars (return bad request in API layer)
- Only allow full-term searches
- Spam Dgraph with requests - one for each possible char filled in as the 3rd char...
- ... 

1. Run legacy api locally (setup by default at localhost:3000)
2. Browse to localhost:3000/items?name=Am
3. Observe the lack of results

katherine-sussol:
Fixes #108. 

## Description

- Adjusted Dgraph query to first query for all unit_of_use nodes, then apply regex as filter (rather than using regex in top-level query). This seems to bypass Dgraph's Regex restrictions. Question also posted to Dgraph forum: https://discuss.dgraph.io/t/error-regular-expression-is-too-wide-ranging-and-can-t-be-executed-efficiently/9254 
- Test added for querying with a 2 char search term (just to validate that some results are returned - as there's already a test for the filtering)

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.

### DIFF
--- a/nx-workspace/apps/api/test/api.test.js
+++ b/nx-workspace/apps/api/test/api.test.js
@@ -173,10 +173,19 @@ describe('Test items endpoint', () => {
     const { code, name } = item;
     const searchName = name.substring(0, 2);
 
+    const searchCodes = items
+      .filter((item) => item.name.startsWith(searchName))
+      .map((item) => item.code);
+
+    const graphResponse = Object.entries(graphResponses.items).reduce((acc, [key, value]) => {
+      if (key != code && searchCodes.includes(key)) acc.data.query.concat(value.data.query);
+      return acc;
+    }, graphResponses.items[code]);
+
     nock('http://localhost:8080')
       .post('/query')
       .query((queryObject) => !!queryObject.timeout)
-      .reply(200, graphResponses.items.all);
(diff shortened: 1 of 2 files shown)


nx-workspace/apps/api/test/api.test.js
```diff
@@ -167,6 +167,38 @@ describe('Test items endpoint', () => {
...
+        method: 'GET',
+        url: `/items?name=${searchName}&exact=false`,
+      })
+      .then((response) => {
+        const { statusCode } = response;
+        expect(statusCode).toBe(200);
+        done();
+      });
+  })
+
```
wlthomson, line 201: wow

PR #122, opened 2020-08-12
#117 nx migrate scripts to tools

### LINKED ISSUE #117: Move docker scripts to tools
Move `auth-service` and `data-service` docker scripts to `tools`.

## Expected Behaviour

Docker scripts should be `tools`, not apps.

## Current Behaviour

Docker scripts are included as apps/in app scope. 

## Possible Solution

- Remove `auth-service` app and move docker scripts to `tools`.
- Move docker scripts from `data-service` to `tools`.

wlthomson:
Fixes #117.

Branched off #121.

## Description

Moves all docker scripts to `tools` directory. Removes `auth-service` app and simplifies `data-service` app.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [x] Tests OK.


wlthomson:
@mark-prins I think this is a good place to stop and consolidate our changes? Thinking main task will be merging the `data` lib classes you've added with the classes I've added to `data-service`?

Current thinking is that it might actually be best to move the sagas, actions, reducers into the `web` app (since they aren't really shared), and extract out the classes I've added (I think it might have actually been my suggestion to put them into a library, so you can blame me)? Reasoning here is that a better strategy might be to keep the redux stuff in the app, but abstract out as much from the reducers etc. as possible into helper classes, types etc. (e.g. keep the actual actions + reducers in the app, but all the logic falls back methods exposed by our classes etc). Would then simplify our library testing too (have no idea how to test sagas/reducers in isolation!).

So, to summarise, that would leave us with the following "blueprint" for the repo:

- 3 apps: `data-service`, `legacy-api`, `web` (not counting `web-e2e` as end-to-end testing is something to leave until we've got a web app with all the `v2` functionality, I think!).
- 2 libs: `ui`, `data` (`ui` for all components, `data` for shared logic, classes, types etc.).
- 2 "toolkits": `dgraph`, `keycloak` (scripts + demo data  for spinning up the containers).

In terms of testing, my current thinking is something like this:

- unit tests for all apps and `data` lib (quite a wide scope, but thinking if we make good use of dependency injection, app tests should be quite limited as the `data` tests will cover most of the logic the apps use).
- no need to tests for the `ui` components (would like to have for more complex components, but thinking we can cover these as part of e2e tests later). 
- in lieu of component tests, would be good to build up storybooks for components as we flesh out the front-end. 

@mark-prins @katherine-sussol Honest feedback? Too much? Not enough? Thoughts?

wlthomson:
@mark-prins I think this is a good place to stop and consolidate our changes? Thinking main task will be merging the `data` lib classes you've added with the classes I've added to `data-service`?

Current thinking is that it might actually be best to move the sagas, actions, reducers into the `web` app (since they aren't really shared), and extract out the classes I've added (I think it might have actually been my suggestion to put them into a library, so you can blame me)? Reasoning here is that a better strategy might be to keep the redux stuff in the app, but abstract out as much from the reducers etc. as possible into helper classes, types etc. (e.g. keep the actual actions + reducers in the app, but all the reducers basically just call methods exposed by our classes etc). Would then simplify our library testing too (have no idea how to test sagas/reducers in isolation!).

So, to summarise, that would leave us with the following monorepo "design":

- 3 apps: `data-service`, `legacy-api`, `web` (not counting `web-e2e` as end-to-end testing is something to leave until we've got a web app with all the `v2` functionality, I think!).
- 2 libs: `ui`, `data` (`ui` for all components, `data` for shared logic, classes, types etc.).
- 2 "toolkits": `dgraph`, `keycloak` (scripts + demo data  for spinning up the containers).

In terms of testing, my current thinking is something like this:

- unit tests for all apps and `data` lib (quite a wide scope, but thinking if we make good use of dependency injection, app tests should be quite limited as the `data` tests will cover most of the logic the apps use).*
- no need to tests for the `ui` components (would like to have for more complex components, but thinking we can cover these as part of e2e tests later). 
- in lieu of component tests, would be good to build up storybooks for components as we flesh out the front-end. 

*e.g. of dependency injection would be the `IdentityProvider` class. In production, `AuthenticationService` etc. would use `KeyCloakIdentityProvider`, but for testing could inject some `MockIdentityProvider` instead, which would make testing these services a lot easier (and then just test `KeyCloakIdentityProvider` in isolation with end-to-end tests or nocks, a la `legacy-api` tests). 

After consolidating, I'm thinking we could have a meeting and figure out how to split out the remaining work, which I think mainly consists of adding tests/storybooks for existing classes/components, and moving forward on the ui front.

@mark-prins @katherine-sussol Honest feedback? Too much? Not enough? Thoughts?

unified-codes/apps/data-service/src/app/AuthenticationService.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export class AuthenticationService {
+  provider: IdentityProvider;
+
+  constructor(provider: IdentityProvider) {
+    this.provider = provider;
+  }
+
+  // TODO: add headers or whatever we need to pass the credentials?
```
wlthomson, line 17: Oops, this `// TODO` is wrong (that stuff should be managed by the identity provider!)

unified-codes/apps/data-service/src/app/AuthorisationService.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export class AuthorisationService {
+  private provider: IdentityProvider;
+
+  constructor(provider: IdentityProvider) {
+    this.provider = provider;
+  }
+
+  // TODO: add permissions, e.g. authorise(user: User, permission: Permission)
```
wlthomson, line 21: Might need to have a discussion how to implement the permission logic? Maybe a `Permission` class and/or enum for roles?

unified-codes/apps/data-service/src/app/User.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class UserError extends Error {
+  constructor(message: string) {
+    super(message);
+  }
+}
+
+export class User {
+  token: JWTToken;
+  roles: string[];
```
wlthomson, line 11: Would perhaps be better that the user just stores token, and we pass the user and operation to `AuthorisationService.authorise(user, operation)`, which then uses an `AccessController` or something to match the task to the roles required, e.g. 

```
  async authorise(user: User, operation: Operation) {
    const isValidUser = this.provider.verifyIdentityToken(user.token);
    if (!isValidUser) return false;
    const isValidOperation = this.controller(user, operation);
    return isValidOperation;
  }
```

Something like that? Think it would be an added bonus that it would be quite easy to test the `AccessController` when its decoupled from the verification of the token.

unified-codes/apps/data-service/src/app/api.ts
```diff
@@ -0,0 +1,41 @@
...
+  const identityProviderConfig = { baseUrl: AUTH_URL };
+  const identityProvider = new KeyCloakIdentityProvider(identityProviderConfig);
+
+  const apolloService = new ApolloService(typeDefs, resolvers, dataSources, identityProvider);
+  const apolloServer = apolloService.getServer();
+
+  return apolloServer;
+};
+
+export const createFastifyServer = (apolloServer, _plugins) => {
```
wlthomson, line 30: IIRC main reason for this is it makes it easier to test, but might need to revist this as couldn't get tests working like they do for the `legacy-api` (might need to directly test the `ApolloServer` instead, which isn't a biggie).

unified-codes/apps/data-service/src/app/resolvers.js
```diff
@@ -29,22 +30,25 @@ const queries = {
...
+    entities: async (_source, _args, context) => {
+      const { dataSources } = context;
       const { type = 'medicinal_product' } = _args?.filter ?? {};
       const query = queries.entities(type);
       const response = await dataSources.dgraph.postQuery(query);
       return response.data.query;
     },
     user: async (_parent, _args, context) => {
-      const { user } = context;
+      const { user, authenticationService, authorisationService } = context;
```
wlthomson, line 49: Idea here is that the two auth services basically represent the source of truth on everyone who exists and everything that they can do, so should provide the resolver with access to everything it needs to decide what can be read/mutated.

unified-codes/apps/legacy-api/src/app/api.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { Server, Route, Options } from './types';
+
+export const createServer = (routes: Route[], opts?: Options) => {
+  const server: Server = fastify(opts);
+  routes.forEach((route) => server.get(route.path, route.opts));
+  return server;
+};
+
+export const createLegacyApiServer = (opts?: Options) => {
```
wlthomson, line 14: Same as above, just makes testing a bit cleaner (although in this instance it might be unnecessary!).

unified-codes/apps/legacy-api/src/app/types.ts
```diff
@@ -0,0 +1,23 @@
+import * as Http from 'http';
```
wlthomson, line 1: We have two fastify servers, so might be useful to put this in the `data` lib to share between `legacy-api` and `data-service`?

unified-codes/tools/scripts/dgraph/setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+echo " * Deleting existing graph data..."
```
wlthomson, line 3: @mark-prins stole your cool `echo` message format 😎.

unified-codes/apps/data-service/src/app/User.ts
mark-prins: agree - the decoupling is important here. nice to have the provider verify the token and then access control is separate.

unified-codes/tools/scripts/keycloak/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo " * Launching keycloak container..."
+docker run -p 9990:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin jboss/keycloak
```
mark-prins, line 4: splitting this out makes it problematic to use - the docker instance doesn't retain data, so the `setup.sh` script always needs to be run.
When split like this - you need to run `run.sh` and `setup.sh`
I see what you're doing, and agree that running and setting up should be different functions. Until we resolve the retention of data when restarting docker instances of keycloak, implementing the separation is just creating a point of failure.

unified-codes/tools/scripts/keycloak/README.md
```diff
@@ -0,0 +1,16 @@
+# Keycloak
+
+Tools for setting up the unified codes [keycloak](https://www.keycloak.org/) service.
+
+## Getting started
+
+1. Install docker [here](https://docs.docker.com/get-docker/).
+2. Run standalone keycloak instance: `chmod +x ./run.sh && ./run.sh`.
```
mark-prins, line 8: nice!

unified-codes/apps/data-service/src/app/api.ts
```diff
@@ -5,34 +5,37 @@ import * as Schema from './schema';
 import * as Data from './data';
 import * as Resolvers from './resolvers';
 
-import ApolloService from "./classes/ApolloService";
-import { KeyCloakIdentityProvider } from './classes/IdentityProvider';
+import ApolloService from './ApolloService';
```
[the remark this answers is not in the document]
mark-prins: I retract that question. all good thanks.

unified-codes/tools/scripts/keycloak/run.sh
wlthomson: My thinking is that in actual production we will probably have to update these scripts anyway, so it might be good to have a consistent script standard? Just IMO, having them separate still allows for `run.sh && setup.sh`, as well as running them separately? Would the possible point of failure (I assume someone accidentally wiping any additional test data?) be a worthwhile trade off for the flexibility/consistency (at least for now?).

unified-codes/tools/scripts/keycloak/README.md
wlthomson: Oops! 

```suggestion
2. Run standalone keycloak instance: `chmod +x ./run.sh && ./setup.sh`.
```

unified-codes/apps/data-service/src/app/IdentityProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  async verifyIdentityToken(token: JWTToken) {
+    const response = await fetch(this.config.baseUrl);
+    const payload = await response.json();
+    const { public_key: publicKey } = payload ?? {};
+
+    if (!publicKey) throw new IdentityProviderError('public key not found');
+
+    const secret = `
```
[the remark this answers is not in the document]
wlthomson: Ah, that sucks. Maybe would be nice to wrap this in a class then, with a `toString` which adds the "padding"?

unified-codes/tools/scripts/keycloak/run.sh
mark-prins: yeah, I think the trade off is worth it. has to be split some time, may as well be now.

unified-codes/apps/data-service/src/app/IdentityProvider.ts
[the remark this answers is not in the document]
mark-prins: actually, this would do the same thing: 

```const secret = `-----BEGIN PUBLIC KEY-----\n${publicKey}\n-----END PUBLIC KEY-----`;```

 That's not too bad is it?

PR #130, opened 2020-08-17
Return explorer results

### LINKED ISSUE #126: Entity Browser not returning results
The Entity Browser component is no longer returning results with Apollo

## Expected Behaviour
Should display results

## Current Behaviour
Shows only `Fetching...` snackbar and no further result

mark-prins:
Fixes #126

## Description
A few issues that had to be worked through
- When using Storybook, the `.env` resolution that nx provides isn't used. To test storybook I've had to include entries in the root `.env` whereas for the web app itself, the entries are in `/apps/web/.env` and `/apps/web/.local.env`
- The changes to authentication meant that anonymous requests resulted in errors. there were a chain of errors involved, hence the number of fixes
- The shape of `dataSources` is different now - should be `dataSources?: () => DataSources<TContext>;`
- The name of the `dgraph` data source had been changed, needed updating in the resolver 

slight off-road: also fixed the circular reference warnings and the compilation errors that were coming up.

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.


unified-codes/apps/data-service/src/app/api.ts
```diff
@@ -11,12 +11,13 @@ import { KeyCloakIdentityProvider } from './IdentityProvider';
 export const createApolloServer = (_typeDefs, _resolvers, _dataSources, _authenticator) => {
   const AUTH_URL = 'http://127.0.0.1:9990/auth/realms/unified-codes';
 
+  const getDataSources = () => ({
+    DgraphDataSource: new Data.DgraphDataSource(),
+    RxNavDataSource: new Data.RxNavDataSource(),
+  });
```
wlthomson, line 17: Hmm, idea is for these to be defaults (but they should be lifted out, actually!). 

In either case, could not get away with:

```
const dataSources = _dataSources ?? { 
  DgraphDataSource: new Data.DgraphDataSource(),
  RxNavDataSource: new RxNavDataSource()
};
```

Just to avoid the inline function?

unified-codes/apps/data-service/src/app/resolvers.js
```diff
@@ -34,15 +34,16 @@ export const resolvers = {
       const { dataSources } = context;
       const { code } = _args;
       const query = queries.entity(code);
-      const response = await dataSources.dgraph.postQuery(query);
+      const response = await dataSources.DgraphDataSource.postQuery(query);
```
wlthomson, line 37: Oops, I think I misnamed these! Should be:

```

File: unified-codes/apps/data-service/src/app/api.ts (hunk 1 above)
Comment (wlthomson) [reply unified-codes/apps/data-service/src/app/api.ts:17]: Also, I wanted to make the imports look pretty, but should probably have imported these as named imports!

```
import * as Schema from './schema';
import * as Resolvers from './resolvers';

import { DgraphDataSource, RxNavDataSource } from './data';
```

File: unified-codes/apps/data-service/src/app/resolvers.js [hunk 3]
```diff
@@ -34,15 +34,16 @@ export const resolvers = {
...
       const [entity] = response.data.query;
       return entity;
     },
     entities: async (_source, _args, context) => {
       const { dataSources } = context;
+      console.warn(dataSources);
```
wlthomson, line 43: surprise

unified-codes/apps/web/stories/EntityBrowser.stories.tsx
```diff
@@ -72,13 +75,16 @@ export const withApolloData = () => {
...
-    );
   }
+
+  return (
+    <Snackbar open={alert.show} autoHideDuration={6000} onClose={hideAlert}>
+      <Alert onClose={hideAlert} severity={alert.severity}>
+        {alert.text}
+      </Alert>
+    </Snackbar>
+  );
```
wlthomson, line 89: Am refactoring this component to include all the alert logic etc... but can resolve this at merge, so come in!

comein

unified-codes/.babelrc
```diff
@@ -1,4 +1,12 @@
 {
   "presets": ["@babel/preset-env"],
-  "plugins": [["@babel/transform-runtime"]]
+  "plugins": [
+    ["@babel/transform-runtime"],
+    [
+      "@babel/plugin-proposal-class-properties",
```
wlthomson, line 6: I just spent 20 mins getting this working! Should have checked...

kiddingme

unified-codes/apps/data-service/src/app/JWT.ts
```diff
@@ -58,19 +58,22 @@ export class JWT {
...
     if (!JWT.validateToken(token)) throw new JsonWebTokenError('jwt malformed');
     return new JWTToken(token);
   }
 
   static parseRequest(request) {
-    const { authorization = {} } = request;
-    return JWT.parseAuthorisation(authorization);
+    const { authorisation = '' } = request;
+
+    return JWT.parseAuthorisation(authorisation);
```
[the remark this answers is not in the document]
mark-prins: In the words of Judith Collins: "what's wrong with being white?!!"

unified-codes/apps/data-service/src/app/User.ts
```diff
@@ -12,6 +12,10 @@ export class User {
 
   constructor(token: JWTToken) {
     this.token = token;
+    if (!token) {
+      this.roles = [];
+      return;
+    }
```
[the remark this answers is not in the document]
mark-prins: the problem is that we have anonymous requests and so there won't be a valid JWT on some requests.
throwing errors just returns no data to the user and that isn't correct when it's an anon request which expects data to be returned. what I'm trying to say in the method is that if a JWT is supplied, validate it and throw an error if it isn't valid. If there isn't one - don't blow up.

which basically says that a `User` can exist without a token. which is how I think it should operate.

Valid concern about early return in ctor. will correct

unified-codes/apps/data-service/src/app/api.ts
mark-prins: > In either case, could not get away with:

no - that's what I tried first. Won't allow it.

unified-codes/apps/data-service/src/app/resolvers.js
mark-prins: oh, ok. will update.

unified-codes/.babelrc
```diff
@@ -1,4 +1,12 @@
...
+      {
+        "loose": true
```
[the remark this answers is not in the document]
mark-prins: results in much terser compiled code (from [babel-plugin-proposal-class-properties](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties))

"However, with { "loose": true }, it will compile using assignment expressions:"

unified-codes/apps/data-service/src/app/api.ts
mark-prins: bah! Tried it now and it does work?!? Did the same thing yesterday and it wouldn't compile 🤷‍♂️

unified-codes/apps/data-service/src/app/resolvers.js
mark-prins: Oopsie

mark-prins:
Thanks @wlthomson - great feedback.
I've addressed with 15277acac

unified-codes/apps/data-service/src/app/api.ts
```diff
@@ -11,12 +11,13 @@ import { KeyCloakIdentityProvider } from './IdentityProvider';
 export const createApolloServer = (_typeDefs, _resolvers, _dataSources, _authenticator) => {
   const AUTH_URL = 'http://127.0.0.1:9990/auth/realms/unified-codes';
 
+  const getDataSources = () => ({
```
mark-prins, line 14: ignore previous comment. the reason that I can't do as you ask is that the `dataSources` has to be a method not an object. Using `??` with the inline method is what fails.

wlthomson: Ah ok! Fair enough, then!

PR #133, opened 2020-08-19
#129 fix broken nx refactor

### LINKED ISSUE #129: Fix broken nx refactor
General issue to complete the merge of the `nx` refactor and get everything running.

## Expected Behaviour

- All projects run.
- All storybooks run.
- All tests run.

## Current Behaviour

TODO.

## Possible Solution

TODO.

wlthomson:
Fixes #129.

NOTE: nearly there, but still a few things to do. Still a draft, but feel free to see if you can get the web app running with a `nx serve web` (note that the user auth and searching functionality is by-passed atm).

## Description

Putting everything together and getting it working!

Including, but not limited to:

- Updating data lib, tidying up/refactoring/migrating classes/types/interfaces.
- Updating ui lib, tidying up/refactoring/migrating components/hooks.
- Refactoring web app, hoisting up state, splitting components into connected/non-connected components, adding state-to-prop + dispatch-to-prop mappers, updating stories etc. etc.
- Adding actions/reducers for user, entities, authenticator, authoriser.
- Fixing `nx` build errors, dependencies, webpack stuff, etc. etc.
- Adding fancy debugging tools, redux toolkit etc.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


unified-codes/libs/data/src/lib/types/JWT.ts
```diff
@@ -57,19 +57,19 @@ export class JWT {
     }
   }
 
-  static parseAuthorisation(authorisation) {
-    if (!authorisation) return;
-    const [_, token] = authorisation.split(' ');
+  static parseAuthorisation(authorisation: string): JWTToken {
+    const [_, jwt] = authorisation.split(' ');
```
mark-prins, line 61: which will break anonymous requests again. At the moment it's all just commented out - shall I fix up in a new issue?

unified-codes/libs/ui/.storybook/webpack.config.js
```diff
@@ -12,5 +12,12 @@ module.exports = async ({ config, mode }) => {
       presets: ['@babel/preset-env', '@babel/preset-react', '@babel/preset-typescript'],
     },
   });
+
+  config.node = {
```
mark-prins, line 16: 3rd time. any way to have it global?

mark-prins (commented):
compiles _and_ runs! 😃 

I have a few comments - nothing major and overall would prefer to have it merged and changes made in much smaller PRs!

wlthomson:
@mark-prins Thanks for review, will definitely try to push this along and make changes in smaller PRs! Just want to get it to a stage where we are all on the same page re. the authentication/user bit before we start all working on it! 

Will update to demo the pattern I'm thinking of for users, as described. If still thinking that it would be better to have `user` for anonymous users, can revisit at that point and make changes if your concerns still not resolved?

unified-codes/apps/web/src/index.tsx
```diff
@@ -1,6 +1,22 @@
...
+import { Provider } from 'react-redux';
 
-import { App } from './App';
+import { ApolloClient, InMemoryCache, ApolloProvider } from '@apollo/client';
 
-ReactDOM.render(<App />, document.getElementById('app'));
+import App from './App';
+import { store } from './store';
+
+const client = new ApolloClient({
```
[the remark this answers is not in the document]
wlthomson: Might be me missing some tidyups! Plan is that the `client` will only be defined here in the main app (might need to duplicate a bit in storybooks though).

P.s. sorry for other "inline" replies, I couldn't reply properly as I had a review in progress without realising!

unified-codes/apps/web/.local.env
```diff
@@ -0,0 +1,4 @@
+NX_AUTHENTICATION_URL=http://127.0.0.1:9990/auth/realms/unified-codes/protocol/openid-connect/token
```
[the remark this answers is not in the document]
mark-prins: yah - was thinking the same thing.
The root `.env` has ended up with all the local bits, so probs ok to encourage devs to copy that to a local verison.

unified-codes/libs/data/src/lib/types/JWT.ts
mark-prins: sounds good - agree, should haven't to authenticate anonymous requests

mark-prins:
@wlthomson one more thing: the `@reduxjs/toolkit` is used in the `store`  but isn't in the `package.json` yet. Had to install in order to build.

unified-codes/apps/data-service/src/app/resolvers.js
```diff
@@ -1,60 +0,0 @@
...
-      return entity;
-    },
-    entities: async (_source, _args, context) => {
-      const { dataSources } = context;
-      const { type = 'medicinal_product' } = _args?.filter ?? {};
-      const query = queries.entities(type);
-      const response = await dataSources.dgraph.postQuery(query);
-      return response.data.query;
-    },
-    user: async (_parent, _args, context) => {
```
mark-prins, line 48: the `User` endpoint was useful for testing auth. will it reappear somewhere else? or removed for good?

wlthomson: Ah, I was a bit confused by the `User` endpoint. My thinking was that the data-service doesn't store anything related to the `User` (it receives a token, but doesn't maintain any state), so I wasn't sure how the frontend was mean't to use this?

At this stage, the idea is the frontend directly grabs the token from keycloak (and then just passes it along with its requests, no graphql types needed). At some stage we might want to add a mutation to `data-service` for the frontend to login (i.e. pass credentials and get the token), but it would remain the case that the `back-end` wouldn't be exposing anything for reading (as it wouldn't be maintaining the user in state, the frontend will just do that with the token). 

Does that make sense or am I missing something?

wlthomson:
> @wlthomson one more thing: the `@reduxjs/toolkit` is used in the `store` but isn't in the `package.json` yet. Had to install in order to build.

Oops! Yeah one of the remaining tasks is to add the react + redux toolkits. Can just do it here, as it's probably not worth a separate PR?

mark-prins:
> Yeah one of the remaining tasks is to add the react + redux toolkits. Can just do it here, as it's probably not worth a separate PR?

yeah, I think jam it in here. no need for a separate PR for it.

unified-codes/apps/data-service/src/app/resolvers.js
mark-prins: It was just parsing the token and returning the user details in a graph. not exactly useful for anyone - though I thought perhaps it might help with the user avatar. It was useful in testing auth, which was really why I put it in. We don't have mutations yet, so no mutation endpoints to secure.

Now that you've done the auth service changes, that helps anyway.

PR #135, opened 2020-08-21
#99 dockerise services

### LINKED ISSUE #99: Dockerise services
As discussed in Sprint 4 meeting (#69) 

## Possible Solution

- Create dockerfile for each separate service in the nx repo (recommended best practice)
  - legacy-api
  - data-service
  - web
- Scripts to individually build docker images
- Scripts to individually run docker images
- Create Docker compose file to start/run all aforementioned services

Future:
- Scripts to run dgraph zero/alpha/ui in separate docker containers (https://dgraph.io/docs/deploy/single-host-setup/ probably enough?)
- Set up docker builds on git push (to master) and publish to dockerhub?

katherine-sussol:
Fixes #99.

## Description

Probably a good place to pause experimentation - have discovered there are many ways of doing the same thing in docker (open to changing! this is just an initial version)
e.g. 
- multistage single dockerfile vs multiple dockerfiles (with inheritance)
- building via `docker build` cmd vs building via `docker compose` vs `nx-docker` extension 
- pushing to [Docker Hub](https://hub.docker.com/) via [nx-docker extension](https://www.npmjs.com/package/@gperdomor/nx-docker) vs [official docker build-push-action](https://github.com/docker/build-push-action)

The idea behind the specific images in this PR is to use them for CI/Deployment/Local Quick Start but not necessarily for debugging/development (as the current setup doesn't support debugging/auto reload). Supposedly it's also possible to do local development exclusively using docker containers (wasn't keen to interfere with current workflow without consultation! Also not sure if worth the investment). 

Some things to test in this version
- Build images (after building services)
  - `unified-codes/unified-codes/tools/scripts/build-dockerfiles.sh`
- Run all services, including dgraph/keycloak (after building images)
  - `unified-codes/unified-codes/tools/scripts/run-all.sh`

**Next Steps**
- Switch from running single standalone Dgraph container to connected Dgraph Zero/Alpha/Ratel containers
- Integrate docker-build-push into GitHub CI workflow (should provide a good base to automate some integration testing?)

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


unified-codes/apps/data-service/src/app/server.js
```diff
@@ -4,7 +4,7 @@ const start = async () => {
   let fastifyServer;
   try {
     fastifyServer = createFastifyServer(createApolloServer());
-    await fastifyServer.listen(4000);
+    await fastifyServer.listen(4000, '0.0.0.0');
```
katherine-sussol, line 7: I feel like I should probably explain this - not having this specified means fastify only listens at 'localhost' or '127.0.0.1' (https://www.fastify.io/docs/latest/Server/) - however when running inside the container requests originate from the docker container's IP address, not localhost

katherine-sussol:
Yeah - docker networking seems to be a fairly large topic on it's own! I think I've only skimmed the surface.
Running via docker compose (i.e `run-all.sh`) sets up a network (at the moment it's just some random name `dockerfiles_default`) and adds the containers defined in the yaml to it - in this network each container can access the services of another by the container name. Ref: https://docs.docker.com/compose/networking/ 

When running them as standalone containers they only join the default bridge network. I haven't actually tested/configured getting them running this way and talking to each other (was doing the end-to end testing via docker compose).

unified-codes/tools/dockerfiles/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+version: '3'
+services:
+  legacy-api:
+    container_name: legacy-api
+    build: 
+      context: ../..
+      dockerfile: tools/dockerfiles/legacy-api.Dockerfile
+    image: unified-codes/legacy-api:latest
+    ports:
+      - "3000:3000"
```
wlthomson, line 10: Looks like you're port mapping here (better way of doing what I said above, probably?)

unified-codes/tools/dockerfiles/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    volumes:
+      - ~/dgraph:/dgraph
+  keycloak:
+    container_name: keycloak
+    image: jboss/keycloak
+    ports: 
+      - "9990:8080"
+    environment:
+      - KEYCLOAK_USER=admin
+      - KEYCLOAK_PASSWORD=admin
```
wlthomson, line 44: Does this retire some of our existing scripts for spinning up these containers?

wlthomson (approved):
LGTM 🎉 ! Structured really nice and covers all the fundamental stuff we need.

Just a few questions/comments, no requested changes from me!

mark-prins (approved):
I really don't know enough about docker to comment.
It works though! and is really nice having all the containers grouped. all runs pretty quickly too, after the initial setup. 
👍

unified-codes/apps/data-service/src/app/data.js
```diff
@@ -3,8 +3,7 @@ import { RESTDataSource } from 'apollo-datasource-rest';
 export class DgraphDataSource extends RESTDataSource {
   constructor() {
     super();
-    // TODO: lift this value out
-    this.baseURL = 'http://localhost:8080';
+    this.baseURL = process.env.NX_DGRAPH_ENDPOINT;
```
mark-prins, line 6: could you do the same for the rx nav endpoint too please? add to `.env` and update here

unified-codes/apps/data-service/src/app/server.js
katherine-sussol: - Hostname - I was thinking the same thing @wlthomson and tried exactly what you mentioned (with 127.0.0.1 explicitly defined) - with this, the request passes through from localhost:4000 -> container but given the container is located inside a docker network, the container itself sees the connection as coming from the docker network rather than 127.0.0.1. I'm sure this behaviour can be customised with a custom network, but I didn't quite want to touch that yet 🙀 
- Port - Good idea to have in `.env` will update

unified-codes/tools/dockerfiles/docker-compose.yml
katherine-sussol: Yeah - it does what you mentioned above internally, I think (though I didn't explicitly define 127.0.0.1 here, it seems to be the default behaviour).

unified-codes/tools/dockerfiles/docker-compose.yml
katherine-sussol: Hmm it could potentially, though in its current state it will also spin up legacy-api and data-service in a non-debuggable way.

Maybe that's fine and we can manually stop them when we want to start debugging? 🤷‍♀️ 
Alternatively we could define a local dev config that just spins up keycloak and dgraph?

mark-prins: +1 for a local dev config, if that's not too hard to do!

katherine-sussol: Cool, might do that in a separate PR! Seems like it shouldn't be too difficult

unified-codes/apps/data-service/src/app/data.js
katherine-sussol: Done with a9661438f6c6f018f3199389168d1e1c2c8a11d0

unified-codes/apps/legacy-api/src/app/server.ts
```diff
@@ -5,7 +5,7 @@ const start = async (): Promise<void> => {
   let server: Server;
   try {
     server = createLegacyApiServer({ logger: true });
-    await server.listen(3000);
+    await server.listen(3000, '0.0.0.0');
```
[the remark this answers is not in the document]
katherine-sussol: Added with c1798ecc6214ffdfdbc2f7c7ca88e707740607e1 and reapplied after merge with 1f87c35c07aeb4b5bdc42839111149e36aff03f6

PR #138, opened 2020-08-21
#137 remove apollo client

### LINKED ISSUE #137: Remove Apollo client from web app
Apollo client is designed to be used to manage local state as well as fetch remote data. Doesn't seem to play nicely with redux, so agreed solution is to migrate away from Apollo client at this stage (can reassess at a later stage if we want to migrate completely to Apollo client, but it doesn't seem worth it right now!).

## Expected Behaviour

All state is managed via redux, all requests and side-effects using sagas.

## Current Behaviour

Data fetching logic is using Apollo client, but also mixed with redux. From a bit of testing and research, this is going to end up pretty messy if we try to scale it!

## Possible Solution

Remove Apollo client, add sagas for fetching from `data-service` (may want to add some helper classes to abstract away the yucky raw `fetch` calls!).

wlthomson:
Fixes #137.

Branches off #139.

## Description

Removes Apollo client and adds redux logic/sagas for handling data fetching.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


mark-prins (approved):
LGTM

mark-prins:
actually a couple of errors. fixed in merge conflict resolution.

PR #156, opened 2020-08-26
#132 fix front end auth

### LINKED ISSUE #132: CORS issue with front end auth
Resolve a problem with keycloak not offering CORS workaround.

## Expected Behaviour
Front end authorisation should work.

## Current Behaviour
Front end authorisation fails with a CORS error.

wlthomson:
Fixes #132 🎉 !

This was a deep rabbit hole, but got there in the end!

To see it in action: 

- Start the keycloak docker (using updated `setup.sh`).
- Run `nx storybook web`.
- Navigate to `Login - With Keycloak On Login Prop`.
- Enter credentials (username: `demo-admin`, password: `pass`).
- On submit, the token string should be displayed 🎉.

## Description

A few things going on here:

- Keycloak will not send a CORs header for a bad request, causing the front-end console to misreport all errors as CORs related (to see what's actually going on, check the keycloak server log).
- Form data was not being correctly formatted, causing keycloak to not receive the request (not sure why, maybe there is another service which forwards requests to keycloak inside the container?).
- Client was not setup with pass (`"asecretthing"`), so attempts to authenticate were rejected.
- Client has `webOrigins` set to a specific localhost port. causing a genuine CORs error if not accessed from that origin. If everything else is setup correctly, this works as expected if you access from the correct origin (but for now easier to just have a wildcard here?).

Off-road (sorry!):

- Simplfied env variable naming (no `NX_`).
- Updated env system (@mark-prins feedback welcome!) to have a simpler way to set envs for debug, production and docker. Local env files are gone (idea being all apps/libraries should be using the same env variables to ensure consensus on hosts, ports, etc.), replaced with root env files: `.env`, `.dist.env` and `.docker.env`. Hopefully this helps!
- Added env variables for base urls and paths (more constants in the code, but more flexible and easier to update ports, paths etc. when testing).
- Updated storybook to latest version (at one point theory was the problem might be storybook end, so updated to make sure we had the latest bug fixes... figured no harm in keeping this in!).

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.

NOTE: PR itself doesn't require any tests, but should now allow us to include keycloak calls in our stories + e2e tests 🎉.


unified-codes/apps/web/src/sagas/AuthenticatorSaga.ts
```diff
@@ -18,10 +18,12 @@ function* authenticate(action: IAuthenticateAction) {
 
   // TODO: should get this from env, as previously!
```
mark-prins, line 19: `// TODO: remove the TODO`

unified-codes/.dist.env
```diff
@@ -0,0 +1,23 @@
+## For production builds, rename this file as .local.env.
```
mark-prins, line 1: Isn't this the production version?

unified-codes/libs/data/src/lib/types/JWT.ts
```diff
@@ -57,22 +57,23 @@ export class JWT {
...
-  static parseResponse(response: any) {
-    const { authorisation = '' } = response;
-    return JWT.parseAuthorisation(authorisation);
+  static async parseResponse(response: any): Promise<JWTToken> {
+    const json = await response.json();
+    const { access_token: jwt } = json;
+    if (!jwt) throw new JsonWebTokenError('access_token malformed');
+    const token = new JWTToken(jwt);
+    if (!JWT.validateToken(token)) throw new JsonWebTokenError('jwt malformed');
+    return token;
```
mark-prins, line 76: :nit It would be nice to lift these two lines out to a shared method

mark-prins (approved):
Looking Good! 
The client pass was the kicker, that's done it for me.
Agree with the .env rearranging - I was running with the nx default concept, prefixing with NX and using local overrides. 
Much neater not to though, like what you've done.

unified-codes/.env
```diff
@@ -1,14 +1,26 @@
...
+DATA_SERVICE_GRAPHQL=graphql
+
+DGRAPH_SERVICE_URL=http://localhost
+DGRAPH_SERVICE_PORT=8080
+
+LEGACY_API_URL=http://localhost
+LEGACY_API_PORT=3000
+
+RXNAV_SERVICE_URL=https://rxnav.nlm.nih.gov
+RXNAV_SERVICE_PORT=8080
```
katherine-sussol, line 25: Just a minor thing - I don't think rxnav exposes their API with ports explicitly, though I'm also kind of now thinking I should have removed the rxnav code when I removed the endpoint which allowed anyone the ability to test it meaningfully 😅

unified-codes/libs/ui/src/lib/components/inputs/organisms/LoginForm.tsx
```diff
@@ -1,12 +1,14 @@
...
+import { IUserCredentials } from '@unified-codes/data';
+
 import Button from '../atoms/Button';
 import Grid from '../../layout/atoms/Grid';
 import LoginInput from '../molecules/LoginInput';
 import Typography from '../../data/atoms/Typography';
 
 export interface LoginFormProps {
-  onSubmit?: (username: string, password: string) => void;
+  onSubmit?: (IUserCredentials) => void;
```
katherine-sussol, line 11: I'm not sure if this is just my dev environment - typescript didn't like this for me and suggested the following change which allowed me to build web:
`onSubmit?: (arg0: IUserCredentials) => void;`

unified-codes/.dist.env
wlthomson: Hm, maybe misunderstanding the Q, but by renaming as `.local.env`, this file will take precedence over `.env` (see https://nx.dev/react/guides/environment-variables#environment-variables). On rereading, I notice the react docs mention having to prepend with `NX`... but the angular docs (the ones I was reading) don't... everything seems to be working, however... so maybe just leave as is (@katherine-sussol?).

In case still not clear, this is how I was thinking:

- Debug: running things locally.
- Production: running things potentially on different hosts (may be local, depending on how we set it up, but may be mixed, e.g. keycloak hosted externally, or something!).
- Docker: somewhat ambiguous, because once we get everything working, docker will be production... but for now, keeping them separate!

It's all quite in flux, was just trying to find a neat way of separating things and keeping everything at the root level... but I think we will probably end up with just a single `.env` with everything local, and then the docker `.env` with everything on the docker network (i.e. at least to start with, the distributed system will be all on a single machine)?

unified-codes/.env
```diff
@@ -1,14 +1,26 @@
...
-NX_AUTHENTICATION_URL=http://localhost:9990/auth/realms/unified-codes
-NX_LEGACY_API_PORT=3000
-NX_DATA_SERVICE=http://localhost:4000/graphql
-NX_DATA_SERVICE_PORT=4000
-NX_DGRAPH_ENDPOINT=http://localhost:8080
-NX_CLIENT_SECRET=asecretthing
-NX_CLIENT_ID=unified-codes-data
-NX_RXNAV_ENDPOINT=https://rxnav.nlm.nih.gov/REST
\ No newline at end of file
+## For local development and storybook usages, rename this file as .env.
```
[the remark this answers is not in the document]
wlthomson: Yes 😉.

Basically, this file should be `.env`, I don't know why I left the rename there (I think initially I named it `.dev.env` or something!). Will remove!

unified-codes/tools/dockerfiles/docker-compose.yml
```diff
@@ -21,11 +20,8 @@ services:
     image: unified-codes/data-service:latest
     ports:
       - "4000:4000"
-    environment:
-      - NX_DGRAPH_ENDPOINT=http://dgraph-standalone:8080
-      - NX_AUTHENTICATION_URL=http://keycloak:9990/auth/realms/unified-codes
-      - NX_DATA_SERVICE_PORT=4000
-      - NX_RXNAV_ENDPOINT=https://rxnav.nlm.nih.gov/REST
+    env_file:
+      - .docker.env
```
katherine-sussol, line 24: Nice - like having the reference to .env over defined in yml itself.

unified-codes/.env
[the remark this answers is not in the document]
wlthomson: So to clarify, this file is always the last env file read, which means that if no other overrides are defined, it will fall back on "debug" mode. If the `.dist.env` file is renamed to be `.local.env`, it will be read first, causing these variables to be ignored.

Yes, no, maybe?

katherine-sussol (approved):
Whoo awesome it works - I also like the `.env` files, particularly the tidy up of hostname vs port.

unified-codes/libs/data/src/lib/types/JWT.ts
wlthomson: Thought that at first too, but thought it might be a bit overkill at this stage? If the pattern comes up again, then definitely agree it should be extracted (rule of three, anyone?).

unified-codes/libs/ui/src/lib/components/inputs/organisms/LoginForm.tsx
wlthomson: Ah yup, that should be `credentials: IUserCredentials`. Will fix!

unified-codes/libs/data/src/lib/types/JWT.ts
mark-prins: is the term `underkill` ever used? 🤔 
all good - leave it as you are.

unified-codes/.dist.env
mark-prins: I see `dist` as production - it's the built / distributable version; the contents of this file looked like production values too. I think that `local` is useful to denote the local dev environment. you have nice values in the example file which goes through source control - called '.prod.env' or '.dist.env' or '.env' and a local version which doesn't live in source control - and which devs copy from the example and tailor to their own needs. My two cents anyway.

if there's any monkeying around to be done with filenames, can we get the build script to do the renaming.

unified-codes/.env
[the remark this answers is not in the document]
mark-prins: sure - I usually have production values in the fallback, and override with local values. Doesn't matter either way really.

unified-codes/.dist.env
katherine-sussol: I prefer the way you've done it @wlthomson  r.e. NX prefix (especially given it works without the prefix), much neater.
R.e. `.env` files in general I like the idea of reserving local for local use that we modify on our machines (it just feels more intuitive in a file with local in its name). Maybe we could keep the current files, but for production, overwrite the `.env` file with the contents of `.dist.env`, and not have any .local file at all? (done via build script of course 😄 ).

Also agree that eventually maybe docker file isn't needed and will just be merged into `dist.env`, but good to wait until we can test a deployment before discarding one or the other.

unified-codes/libs/data/src/lib/types/JWT.ts
katherine-sussol: Hm, would `underkill` be akin to non-working software or really badly written software? Would it be correct to say @wlthomson has applied lethal force to the problem? I'm not really adding anything here, other than expressing agreement with the rule of three 😸

unified-codes/.dist.env
wlthomson: @mark-prins Maybe I was unclear again 😨! That file is indeed the production file, renaming as `.local` just gives it priority over the other `.env` file (which is debug). Does that make sense?

Yeah, I agree, I initially had it so all files were prepended, e.g.: `.debug.env`, `.dist.env`... and the user would rename whichever one was appropriate to `.env`. Does that sound about right?

unified-codes/apps/web/src/sagas/AuthenticatorSaga.ts
wlthomson: 0c8d169df.

unified-codes/libs/ui/src/lib/components/inputs/organisms/LoginForm.tsx
wlthomson: 9bca63665f07d2bddfb19cb2696a9f064b689bee.

unified-codes/.dist.env
wlthomson: ef03c25e3.

PR #162, opened 2020-09-02
#155 create ui entity type toggle component

### LINKED ISSUE #155: Create component(s) for toggling entity types to display
## Expected Behaviour

From invision 
Screen Shot 2020-08-25 at 3 33 07 PM

Picking something small to try to get started with this sprint. I don't really know if we can distinguish between a Drug and Product yet - but can still probably create some buttons/toggles and get familiar with the app!

katherine-sussol:
Fixes #155 .

## Description

Small PR (to check if on the right track) - creating icons and basic toggle button.

Haven't put it into the UI yet - one thing I wasn't too sure about is styling conventions, seems like the convention is to keep most of the styling out of the lib components and leave it to the web app. However as bg colour is dependent on selection property I put it here. Let me know if any of my understanding above is wrong 😄 

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.


unified-codes/libs/ui/src/lib/components/inputs/molecules/ToggleButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+import CheckCircleIcon from '../../icons/atoms/CheckCircleIcon';
+
+const styles = {
+  toggleButtonPrimary: {
+    backgroundColor: '#2B83A1',
+  },
+  toggleButtonSecondary: {
+    backgroundColor: '#5CCDF4',
+  },
+};
```
mark-prins, line 14: I wondered about shifting the colours to the theme. 
Turns out that it is a bit of work and requires updating this file, the `muiTheme` and the `ToggleButtonStories.tsx` as well. 🙄 

The suggestion didn't show the diff very well - I could commit changes to show what I mean instead, if you like. Or we just ignore, which is fine. What I've done is nice and it means that the colours can be reused elsewhere. Aside from that.. no big benefit.

mark-prins (approved):
looks good!

unified-codes/libs/ui/src/lib/components/inputs/molecules/ToggleButton.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+import ToggleButton from './ToggleButton';
+
+export default {
+  component: ToggleButton,
+  title: 'ToggleButton',
+};
+
+export const Unselected = () => {
+  return <ToggleButton>Toggle Button</ToggleButton>;
+};
```
mark-prins, line 12:
```suggestion
export const Unselected = () => (
  <ToggleButton>Toggle Button</ToggleButton>
);
```

an alternative, with a few less characters required

katherine-sussol: Switched it over with 1169b59, nice to save those chars 😸

unified-codes/libs/ui/src/lib/components/inputs/molecules/ToggleButton.tsx
katherine-sussol: Ah, I'd be interested to see! Sounds useful

mark-prins: 965ac822c with the addition of the `ThemeProvider` in the story only required because the theme isn't provided automatically in the storybook context.

In the web, that just works ( it is passed down in `index.tsx` )

katherine-sussol: Nice!

PR #176, opened 2020-09-10
#163 toggle button menu

### LINKED ISSUE #163: Add toggle button menu to web UI
Follow on from #155 (which just created the toggle buttons) to group the buttons into some kind of 'toggle menu' component and insert under the header.

## Expected Behaviour

Buttons in a group added to the main UI
Screen Shot 2020-09-02 at 4 43 07 PM
## Current Behaviour

No buttons in the UI :(

katherine-sussol:
Fixes #163.

## Description

An attempt to create the entity type filter menu (decided to keep the PR just creating the component and next step for this particular issue will need to be a separate issue/PR to wire it up to filter the explorer). Took a stab at styling it, feedback appreciated (should the styles be removed from the toggle button menu group?)

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


unified-codes/apps/web/src/components/molecules/EntityTypeFilter.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { ToggleButtonGroup } from '@unified-codes/ui';
+
+export type EntityTypeFilter = React.FunctionComponent;
+
+export const EntityTypeFilter: EntityTypeFilter = () => {
+  const [entityStates, entityStateChange] = React.useState([
+    { name: 'Drugs', active: true },
```
mark-prins, line 8: I was wondering how to make use of the filter, given that all of the items we have are `medicinal_product` and thinking that someone needs to review the current dataset and codify items, as `drug` and `other`

unified-codes/apps/web/src/components/molecules/EntityTypeFilter.tsx
```diff
@@ -0,0 +1,30 @@
...
+    { name: 'Unit of Use', active: false },
```
mark-prins, line 9: `Drugs` is plural whereas `Unit of Use` and `Other` are singular - we need to decide what to have here I think
Also - `Drug` is top level in the hierarchy and `Unit of Use` is much lower down - given that we only show top level items, probably doesn't make sense to have both together in a filter.

Not a coding comment, but something that's been troubling me. any ideas how we manage this?

unified-codes/libs/ui/src/lib/components/inputs/molecules/ToggleButtonGroup.stories.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import { ThemeProvider } from '@material-ui/core/styles';
+import theme from '../../../../../../../apps/web/src/muiTheme';
```
mark-prins, line 3: I've done something like this elsewhere and expected someone to ping me for it in a code review 😉  Not sure the best way to make this more elegant. will think on it.

mark-prins (approved):
I like that a lot, nice work. The integration with the theme is spot on.
My comments are mainly around data rather than coding, and reflect my own uncertainties. Not something to hold up the PR though, they're more general concerns to be worked through.

unified-codes/apps/web/src/components/molecules/EntityTypeFilter.tsx
katherine-sussol: Yeah, I was thinking `other_products` might be suited as another node/entity type that has all the non-drug things directly linked as children in Dgraph? That way we'd be able to pull all 'other' and children when the other tab is selected.

unified-codes/apps/web/src/components/molecules/EntityTypeFilter.tsx
katherine-sussol: Hmm, I've also been a little unsure about the use case. I was also thinking it doesn't make sense to have both, but was thinking maybe people would be more interested in seeing unit of uses in the table rather than the medicinal product level? But if someone clicked through on a specific product they'd then be able to browse the hierarchy? Or something to that effect...? 🤷‍♀️

PR #173, opened 2020-09-10
#152 apply styling to explorer

mark-prins:
Fixes #152 

## Description
Makes the table look more like the mock up. Could be more beautiful, but then, couldn't we all?
- Has the table height more adjusted to the max available in the body area, with a scroll for the results. 
- UC icon updated with the version from Pete.
- header and footer heights have been reduce a little
- footer changed from `Drawer` to `Paper` as the drawer causes a few issues
- `Drawer` removed from project ( no longer used )
- theme updated to include some customised options
- background colour of page updated to give contrast to table

### Documentation:

Tick below if one of the following applies:

1. Documentation has been added to cover changes introduced by this PR.
2. This PR requires no changes to documentation.

- [ ] Documentation OK.

### Tests:

Tick below if one of the following applies:

1. Tests have been added to cover changes introduced by this PR.
2. This PR requires no new tests.

- [ ] Tests OK.

PR #177, opened 2020-09-11
#174 add ordering of results to explorer

### LINKED ISSUE #174: Add ordering of results to explorer
Currently results are returned in database default order.
I really like alphabetical. Would be nice to have sort options in UI too

## Expected Behaviour
Results should be ordered by default in a meaningful way.
User should have the ability to order as required.

1.
2.
3.
4.

mark-prins:
Fixes #174 

## Description
Adds the ability to sort results by any of the columns shown, ascending or descending.
Shows a caret / arrow next to the column header to indicate which is sorted.
The sort itself is done in dgraph, with the options passed back by the `data-service`. currently searching only one field, haven't explored multi-field searching.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


mark-prins:
Oh, this branched off #173 - that one should be merged first to make review of this one easier.

unified-codes/apps/web/src/components/molecules/Footer.tsx
```diff
@@ -1,25 +1,37 @@
...
+const getStyles = (theme: ITheme) => ({
+  root: {
+    height: 120,
+    position: 'fixed' as position,
+    width: '100%',
+    bottom: 0,
+    alignItems: 'center',
+    backgroundColor: theme.palette.background.footer,
+    display: 'flex',
+    flexDirection: 'column' as flexDirection,
```
wlthomson, line 23: I was initially hoping to try to avoid components having baked in `flex` styles and only use the material UI grid API: https://material-ui.com/api/grid/. My thinking is it could get messy if we mix `Grid` flexbox with `display: 'flex'` etc. Is there something we can't do here using the grid layout system, or just what was familiar?

unified-codes/apps/web/src/components/molecules/Footer.tsx
```diff
@@ -43,10 +55,9 @@ const getStyles = (theme: Theme) => ({
 
 export type Footer = React.FunctionComponent<FooterProps>;
 
-const StyledDrawer = withStyles(getStyles)(Drawer);
 const _Footer: Footer = ({ classes }) => {
   return (
-    <StyledDrawer anchor="bottom" variant="permanent">
+    <Paper className={classes.root} square elevation={0}>
```
wlthomson, line 60: At this stage premature, but what do you think about `elevation={0}` etc. being something that is imported rather than specified inline? Will it be shared between `<Paper>` components or differ throughout the app?

unified-codes/apps/web/src/components/templates/Explorer.tsx
```diff
@@ -14,7 +14,13 @@ import { ExplorerActions } from '../../actions';
 import { IExplorerData, IState } from '../../types';
 import { ExplorerSelectors } from '../../selectors';
 
+import { withStyles } from '@material-ui/core/styles';
+import { ITheme, overflow } from '../../muiTheme';
+
+const FOOTER_HEADER_HEIGHT = 300;
+
 export interface ExplorerProps {
+  classes?: IEntityBrowserClasses;
```
wlthomson, line 23: Bit confused about prop drilling classes. Could this be something defined on a per component basis rather than passed down?

unified-codes/apps/web/src/components/templates/Explorer.tsx
```diff
@@ -23,9 +29,31 @@ export interface ExplorerProps {
...
+      marginTop: 5,
+      maxHeight: `calc(100vh - ${FOOTER_HEADER_HEIGHT}px)`,
+      overflowY: 'scroll' as overflow,
+      '& th': { backgroundColor: theme.palette.background.toolbar, fontWeight: 700 },
+      '& thead > tr': { borderBottom: borderStyle },
+      '& tr > th': { borderRight: borderStyle },
+      '& tr > td': { borderRight: borderStyle, borderBottom: 0 },
+      '& tr > td:last-child': { borderRight: 0 },
+      '& tr > th:last-child': { borderRight: 0 },
+      '& tr > td:first-child': { fontWeight: 700 },
```
wlthomson, line 48: Thinking some of this could potentially be abstracted to a shared styles file/folder? Would you be thinking to refactor away these hardcoded style constants and stuff?

unified-codes/apps/web/src/components/templates/Explorer.tsx
```diff
@@ -23,9 +29,31 @@ export interface ExplorerProps {
...
+    },
+  };
+};
+
 export type Explorer = React.FunctionComponent<ExplorerProps>;
 
 export const ExplorerComponent: Explorer = ({
+  classes,
```
wlthomson, line 56: As above, not too sure about prop drilling `classes` via the parent to the child when the styles can be imported directly (would this not end up with parents having to manage a really large number of `classes` as the tree got deeper?)

unified-codes/apps/web/src/components/templates/Explorer.tsx
```diff
@@ -59,17 +87,28 @@ export const ExplorerComponent: Explorer = ({
...
-        entities={entityData}
-        onChangePage={handleChangePage}
-        onChangeRowsPerPage={handleChangeRowsPerPage}
-        onClear={handleClear}
-        onSearch={handleSearch}
-        variables={variables}
-      />
-    </Grid>
+    <EntityBrowser
+      childProps={childProps}
```
wlthomson, line 102: Think same issue as `classes` here, I think? Not sure about the parent having a generic prop which handles all the child props, do you think could be better to figure out what props we need to drill through and make them explicit?

unified-codes/apps/web/src/components/templates/Header.tsx
```diff
@@ -51,7 +51,7 @@ export type Header = React.FunctionComponent<HeaderProps>;
 
 const _Header: Header = ({ classes }) => {
   return (
-    <AppBar position="static" className={classes.root}>
+    <AppBar position="fixed" className={classes.root}>
       <Grid item className={classes.title1}>
```
wlthomson, line 55: I'm thinking if `Header` knows it needs `classes.title1`, would we lose anything by directly importing the class in the `Header` component file?

wlthomson (approved):
A few architectural things, mostly stuff that can be handled by refactors, I think, but would be good to get your thoughts, anyway!

Going to approve anyways, so as to not block stuff. Can make any agreed changes/refactors in new issues where needed, that way hopefully won't hold you up!

wlthomson:
@mark-prins I lost connectivity and had to redo the review... but looks like some stuff ended up getting remembered by github, so feel free to ignore any duplicates!

unified-codes/apps/data-service/src/app/resolvers.ts
```diff
@@ -39,11 +39,12 @@ export const resolvers = {
...
 
-      const { type = 'medicinal_product', code, description } = filter ?? {};
+      const { type = 'medicinal_product', code, description, orderDesc, orderBy } = filter ?? {};
+      const order = `order${orderDesc ? 'desc' : 'asc'}: ${orderBy}`;
       const query =
         description && description.length
-          ? queries.entitiesByDescriptionAndType(type, description)
-          : queries.entitiesByType(type);
+          ? queries.entitiesByDescriptionAndType(type, description, order)
+          : queries.entitiesByType(type, order);
```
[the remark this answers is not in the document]
mark-prins: you are quite right!
I refactored in #178 - it was bugging me, and I was needing to add another param.
Of course, a bit ol' blob of JSON might have worked? or an object?

unified-codes/apps/data-service/src/app/schema.ts
```diff
@@ -35,6 +35,10 @@ const typeDefs = gql`
     description: String
     "Search by Level in Product Hierarchy"
     type: String
+    "Field to order by"
+    orderBy: String
+    "Sort ascending"
+    orderDesc: Boolean
```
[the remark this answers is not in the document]
mark-prins: nice! thanks. I'll pinch that.. da9e29f0b

unified-codes/apps/web/src/App.tsx
```diff
@@ -24,7 +24,10 @@ const getStyles = (theme: Theme) => ({
     paddingRight: 0,
   },
   content: {
-    marginTop: 24,
+    backgroundColor: theme.palette.background.paper,
+    marginTop: 96,
+    maxHeight: '100vh',
+    paddingBottom: 10,
```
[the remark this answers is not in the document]
mark-prins: like a big stylesheet file? opinions divided on where to place styles - next to usage in the component or somewhere outside. I like the idea of lifting them out, will have to research and discuss.

unified-codes/apps/web/src/components/molecules/Footer.tsx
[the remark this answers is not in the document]
mark-prins: sure- maybe raise an issue though. that'll take a day of exploration.

unified-codes/apps/web/src/components/templates/Explorer.tsx
mark-prins: yes, except that `EntityBrowser` is wrapping so many components and I want the styles in the web, which is why I passed down in classes. another discussion point?

unified-codes/apps/web/src/components/templates/Explorer.tsx
mark-prins: oh, ok. I see where you're heading. I wouldn't want library components to be importing styles though. could I import at the web level and pass in to the lower components?

wlthomson: I agree with not importing into library components (as that would just cause coupling as you imply), but since `Explorer` is a locally defined component, passing in `classes` from another locally defined component seems like potentially redundant prop drilling (but I'm happy to be wrong if this is an accepted pattern). 

I think its fair enough to have a generic `classes` and `style` prop for all components, and for parent classes to drill classes down the tree. But I think if we go this route we should include it in all our library components, so it's a consistent API.

Having said that, I think it's best to avoid relying too much on drilling generic props, mainly because I think it encourages coupling between components and could get messy quite quickly.

I think a good option is to have `classes` and `styles` for all components, but only for presentational stuff, and as much as possible only for the component itself, rather than specifying all the child styles in the parent. I think a better way to handle the issue of styling the subcomponents is to expose them as props directly, i.e. library components containing other library components expose their children as optional props. Since the children are also exported by the library, you can then wrap them in a closure with all the props you want them to get (again, though, ideally only presentational stuff, since their designed to be "dumb"). I've seen this pattern in a few places, and seems to have similar benefits as OOP dependency injection.

Here's an example with styling the header for an `EntityTable`:

```
const HeaderComponent = React.useCallback((props) => <EntityTableHeader {...props} style={YOUR_CODE_HERE}/>
...

<EntityTable
  headerComponent={HeaderComponent}
  rowComponent={RowComponent}
  ...
/>
```

I think this would also tidy up the library components a bit, which I think have grown to include conditional logic which I think we generally on the same page with wanting to avoid as much as possible (i.e. make the component as "dumb" as possible, push conditions up the tree as much as possible)?

wlthomson:
@mark-prins Looks like this is merged, do you want me to raise issues for the discussion above?

mark-prins:
yeah, thanks. It was sounding like a new issue to me, due to the scope of the changes. 
Would be good to discuss how we handle the styling, maybe you'd have some time to research a good way to do it?

unified-codes/apps/web/src/components/templates/Explorer.tsx
wlthomson: Yup, understand your point. Something to be discussed in a dedicated issue, as you say.

PR #178, opened 2020-09-14
#175 refactor entities redux shape

### LINKED ISSUE #175: Refactor entities redux shape
Would like to refactor the storage of entities in the `store` as a class rather than the current POJO which was an extension of the array of entities.

## Expected Behaviour
Provide the entities, along with
- total number of results
- methods for fetching the next/previous pages of results?

## Current Behaviour
Current shape is this:
```
  const data: IExplorerData = {
    data: entities,
    hasMore: false,
    totalResults: 0,
  };
```

## Possible Solution
Create an `EntityCollection` class which provides
```
{
  count: number;
  data: Array<IEntity>;
  totalCount: number;
}
```

mark-prins:
Fixes #175 

## Description
Moves the pagination out of the `data-service`! 😃 
this is now handled by dgraph, as well as sorting and searching.

Have refactored as per @wlthomson's request in #169  and the result is much nicer (thanks!)

**Offroad confessions:** 
- added an Enter key handler for the search input, as this has bugged me
- changed the index type for `type` to be `term` as this allows filtering by multiple types, which then supports the toggle buttons ( just send multiple types as a string, space separated e.g. `"medicinal_product unit_of_use"`

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


mark-prins:
ready for review again - have merged master in and fixed up errors
the style is still out - have resolved that in #179

unified-codes/apps/data-service/src/app/resolvers.ts
```diff
@@ -41,18 +40,11 @@ export const resolvers = {
 
       const { type = 'medicinal_product', code, description, orderBy } = filter ?? {};
       const order = `order${orderBy.descending ? 'desc' : 'asc'}: ${orderBy.field}`;
```
katherine-sussol, line 42: Sorry, I know this is from an older PR - but from my prodding of the API, I think it may be nice for orderBy (and the offset/first fields) to either have some default values or to be made mandatory in the schema? I think as it is it's fine for our use case in UI though so doesn't need to be done now necessarily.

katherine-sussol (commented):
Looks good - the collectioning/paging/searching seems like a reasonably fiddly piece of work (it's always fun pointing out issues when someone else has done all of the hard work - sorry for being that person 😅 ). Just a couple of comments!

unified-codes/apps/data-service/src/app/schema.ts
```diff
@@ -19,13 +20,12 @@ const typeDefs = gql`
     "Request an entity by code"
     entity(code: String!): Entity
     "Request all entities with optional filter - Default behaviour: return all medicinal_products"
-    entities(filter: EntityFilterInput, first: Int, offset: Int): EntitiesResponse
+    entities(filter: EntityFilterInput, first: Int, offset: Int): EntityCollection
   }
 
-  type EntitiesResponse {
+  type IEntityCollection {
```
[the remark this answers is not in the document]
mark-prins: yes, quite right. I fixed in another branch and didn't know where it came from! thanks :) 
8a0fdceb6

unified-codes/apps/data-service/src/app/resolvers.ts
mark-prins: yes, thanks!
The classes (`EntitySearchFilter` and now `EntitySortInput`) have defaults which covers current usage - I didn't make them mandatory in the schema though, you think I should?

katherine-sussol: I'd say yes just because when they're missing it makes the data-service throw out some not so user-friendly messages (which is kind of fine for us and the web app - but maybe not so much for third parties/other applications later down the line?).

katherine-sussol (approved):
👍  looks good - trying not to hold up the PR Train

unified-codes/apps/data-service/src/app/resolvers.ts
mark-prins: good point! thanks.

PR #184, opened 2020-09-16
#183 create Entity mock data

### LINKED ISSUE #183: Add mock data to Explorer story
When viewing the web storybook, the `Explorer` component has no data.

## Possible Solution
Add the current mock data to a mock data file and use for all cases where mock entity data is required.

mark-prins:
Fixes #183 

## Description
Have moved the existing two instances of mocks (`EntityTableRow` and `EntityBrowser`) into a shared file and used it in those two, as well as `Explorer`

Side fix: the number of pages in the pagination component was not correct ( should default to `0` not `1` ) and the `EntityCollection` class was not taking `TotalLength` from the supplied array, if not specified in the params.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


katherine-sussol (approved):
👀 s good to me!

unified-codes/libs/data/src/lib/mocks/Entity.ts
```diff
@@ -0,0 +1,19 @@
...
+    code: 'a1004adb',
+    description: 'Metronidazole',
+    type: 'medicinal_product',
+  },
+  {
+    code: 'GH89P98W',
+    description: 'Paracetamol',
+    type: 'medicinal_product',
+  },
+];
```
wlthomson, line 19: Relatively long comment coming, but not much work required (promise)!

Understand the motivation here, but going to be annoying again and stick my 👃 in and suggest we keep mock data as WET as possible. That means some duplication, but I think it's the best approach. I probably should have discussed this earlier, not great communication on my part (but getting it early!).

Thought about this a bit while doing restful service jest tests. After discussing pretty extensively with @joshxg, I refactored from something like what we've got here (shared mock data), to having everything is in each test cases (see https://github.com/openmsupply/unified-codes/blob/master/unified-codes/apps/legacy-api/src/app/api.spec.js).

Few reasons for this:

1. Each story/test case is a single "unit" with no dependencies except for the code being tested. I.e. everything needed by a test is declared in the test, no external dependencies (I think of this as each test only having one variable, which is the code being tested, no data/code outside the test which can influence if the test passes or not).
2. A test can be understood by just reading the test, no cross-referencing with mock data files (which can get big).
3. Shared data encourages using the same data for multiple tests, which can cause false positives due to not enough variation in the data.

I think it holds doubly in this case, because the mock data is being put into the library. IMO this couples the library to the implementation too tightly (ideally web storybooks shouldn't need to know anything about the library implementation except for the component props/interfaces etc.). 

What you think about:

1. Keep mock data for the web app in the web app.
2. No mock files, keep data DRY and declared in test cases (rather than imported/injected).

This is also something I had to learn, as it's tempting to try to bundle the shared code in one place, but I think it's a bit of a smell in this case (some people go as far as to say all mocks are smells!). This is probably something we want to do organisation wise (I think I can speak on @joshxg behalf that he agrees on this?).

wlthomson (changes requested):
Sorry! Wouldn't request changes if wasn't a relatively quick fix (and something that's probably important to agree on earlier rather than later!).

unified-codes/libs/data/src/lib/mocks/Entity.ts
wlthomson: @joshxg Sorry for getting you involved, just ccing because of our discussion about this recently! If I'm misrepresenting your thoughts, feel free to correct me, though!

joshxg: HI!

I won't pretend I know what I'm reading/looking at in regard to the project specific details - but my thoughts currently on shared mock data are that they save you many seconds when writing a test from not having to copy paste but can cause many more painful ones when some unit test somewhere accidentally mutates data or some data is changed somewhere to suit some unit test somewhere. 

Of course the usual WET v DRY tradeoffs come into play - maintenance can take longer, changing implementation (😨 ) can take longer, actually writing tests can take longer (all arguable!) - I'm not convinced that either end of the spectrum (none ever v yes always) is right, but I know that so far, for me, sharing mock data has only caused me issues and not saved me much time!

mark-prins: thanks guys - agree, I've also had issues with shared and also with mutated mock data. I'll wind it back.

wlthomson (approved):
LGTM! 🎉

Issue #191, opened 2020-09-19
Refactor web redux shape


wlthomson:
The web redux shape includes classes, which is "highly encouraged" to avoid as it breaks some core redux functionality (see https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state). I think may have been some confusion re. the `EntityCollection` class (IIRC intention was to use in backend and as a helper in reducers etc. but not to be stored directly in state).

## Expected Behaviour

Store should not include any class instance (POJOs OK). 

## Current Behaviour

Store contains class instances.

## Possible Solution

Refactor `EntityCollection` out of the store and into reducers etc.

PR #192, opened 2020-09-21
#185 search results page

### LINKED ISSUE #185: Start SearchResults page
## Expected Behaviour

Users should be able to find out more about medicinal products they find in the explorer (good way to show off the external codes too)

## Possible Solution

- Turn items in Explorer into link to a results page 
- On click, fire entity query requesting all children and properties of the selected medicinal_product
- Dump the data into a dummy results page (can deal with layout/styling later - I think already enough things for one issue)

katherine-sussol:
Fixes #185.

## Description

Apologise in advance for the ugliness of this new page - just thought it better to get feedback early before potentially going further down the current track 🛤️ . Happy to adjust as needed (with a promise as the next step to make the page render something resembling a proper UI and not just a JSON string).

Some questions/feedback requested:
- Search vs Browse for the different views. Thinking 'Browse' is a more appropriate term for the details view (because a user is 'browsing' a particular product), and search is more appropriate for the explorer (as they're doing search-like things there)? Not really that fussed but struggled a bit trying to name the template with some variant of search - went with PropertyBrowser in the end 🤷 
- Redux ... when to use. Should the Entity go in the store? This seems to be a more 'one-off' type of view and it seems like a lot of extra 'stuff' needed to use redux which might be overkill? Or is there still value in it for consistency with other API calls and code structure? 

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


mark-prins:
@katherine-sussol - thanks 🙇‍♂️  I agree - better to get something in and working. appreciate the feedback request. Also good to have smaller PRs to look through.

> Search vs Browse for the different views. 

Yes! I have the same question. I think that the distinction arose from the designs, rather than functionality - and we have merged the two concepts. Shall we pick one and run with it? I'd suggest 'Browse` - to me, what we have is a browse & filter atm

> Redux ... when to use. Should the Entity go in the store?

I don't think that it's necessary to pop in the redux store for this one. I use the store for data that can change, or requires async fetching, or for communication across components - none of which really affect this page. We may need a loading spinner while fetching data, other than that - it's all just one set of data.

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityTableRow.tsx
```diff
@@ -8,14 +8,15 @@ export interface EntityTableRowProps {
...
 }
 
 export type EntityTableRow = React.FunctionComponent<EntityTableRowProps>;
 
-export const EntityTableRow: EntityTableRow = ({ rowProps, cellProps, entity }) => {
+export const EntityTableRow: EntityTableRow = ({ rowProps, cellProps, entity, onEntitySelect }) => {
   const { code, description, type } = entity || {};
   return (
-    <TableRow {...rowProps}>
+    <TableRow {...rowProps} onClick={() => onEntitySelect(code ?? '')}>
```
mark-prins, line 19: it's not likely ( or ever going to happen ) - perhaps don't do anything if the code is undefined?

unified-codes/apps/web/src/components/templates/PropertyBrowser.tsx
```diff
@@ -0,0 +1,65 @@
...
+          code
+          description
+          type
+        }
+      }
+    }
+  }`;
+
+  const getEntity = async (url: string) => {
+    const response = await fetch(url, {
```
mark-prins, line 44: this is all good - and I know that I said in my earlier comment to keep it local and not use redux..
On looking at it, I think it would be nice to use the redux saga and keep the `fetch` out of the component.

That way, you can dispatch an action in `getEntity` and just connect an `Entity` from the store to this component.

Either approach is fine - I like the idea of keeping the fetching in the same place, which would mean moving to the saga. and it simplifies this component, so that it is only concerned with display of data.

mark-prins (approved):
That looks great, thanks Kat!
I have some suggestions - only some of which are contradictory. Ok with it going in and then refactoring, or modifying this PR, whichever works for you.

unified-codes/apps/web/src/App.tsx
```diff
@@ -9,7 +9,7 @@ import { Container, Grid, AlertBar } from '@unified-codes/ui';
 import { IAlert } from '@unified-codes/data';
 
 import { AlertActions } from './actions';
-import { Explorer, Footer, Header, Login } from './components';
+import { Explorer, Footer, Header, Login, PropertyBrowser } from './components';
```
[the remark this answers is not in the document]
katherine-sussol: Indeed! Thanks for the suggestions. I agree, `Details` would probably work to match the URL. The reason I didn't go with Entity in the name was due to potential confusion with the EntityBrowser and EntityTable components 😖, but since they are in a different folder maybe they would have also been ok 😄

unified-codes/apps/web/src/components/templates/Header.tsx
```diff
@@ -63,9 +63,9 @@ const _Header: Header = ({ classes }) => {
       </Grid>
       <Grid item style={{ flex: 1 }}>
         <Grid container className={classes.menu}>
-          <Grid item>
+        <Grid item>
             <Link component={RouterLink} to="/">
-              Browse
+              Search
```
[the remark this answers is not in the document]
katherine-sussol: I think I'll leave it as it was previously (`Browse`) since it's not really in the context of this issue anymore (originally tried to link directly to the details page from the header and realised it didn't make sense).

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityTableRow.tsx
katherine-sussol: Hm, true. That would be better than redirecting to details with no code (if somehow it did happen).

unified-codes/apps/web/src/components/templates/PropertyBrowser.tsx
katherine-sussol: Hehe, yeah I was very unsure about which was better. I guess it may be a good opportunity to fiddle with redux so I may move it for that reason as well as the benefit of a smaller component 👍

katherine-sussol:
Have added details with a separate set of actions (following existing convention). Debated a bit whether or not it should be considered part of the explorer or not (it's still exploring a product? 🤷). I can move if it's problematic though 😅 

May have also broken the rule about class instances in redux (just saw issue #191). What's better here? Just a generic JS object?

unified-codes/apps/web/src/reducers/DetailsReducer.ts
```diff
@@ -0,0 +1,38 @@
...
+): IDetailsState => {
+  const { type } = action;
+
+  switch (type) {
+    case DETAILS_ACTIONS.FETCH_ENTITY: {
+      return { ...state, loading: true };
+    }
+    case DETAILS_ACTIONS.FETCH_SUCCESS: {
+      const { entity } = action as IDetailsFetchSuccessAction;
+      return { ...state, entity, loading: false };
```
mark-prins, line 27: not a problem at the moment, and this isn't exactly going to happen. but..
I'd recommend clearing the error here as well. in case
- fetching
- fetch failure ( sets `error` )
- fetching 
- fetch success ( `error` is still set )

this is only an issue if error is used to display a problem, and that isn't happening now. nor is likely to.
and I can't see the need to fetch, fail, re-fetch. 

so, this is more abstract than useful.

mark-prins (approved):
nice! 😃  
I like the separation of getting data and displaying data and it makes the component much smaller

unified-codes/apps/web/src/reducers/DetailsReducer.ts
katherine-sussol: Thanks for the suggestion 😸  have applied in ca31cc6

mark-prins:
> May have also broken the rule about class instances in redux...

No, I think you have it right. At least as I understand it

katherine-sussol:
😅  ok - have left as is for now

PR #204, opened 2020-09-28
#198 layout details page

### LINKED ISSUE #198: Layout for details page
## Expected Behaviour

Human-readable descriptions of products 

## Current Behaviour

Machine-readable descriptions of products

## Possible Solution

Expand current `<div>` tag to something more substantial (maybe loosely follow the inVision mockup).

katherine-sussol:
Fixes #198. 

## Description

- Adds some form to the details page, including children, properties and external link to NZULM as example
- Adds spinner when fetching
- Does not add good taste or style to the page (designer to review and prettify?) 

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


unified-codes/apps/web/src/components/templates/Details.tsx
```diff
@@ -2,41 +2,107 @@ import * as React from 'react';
...
+          type: property.type,
+          value: property.value,
+        });
+        break;
+    }
+  });
+
+  return (
+    <Grid container className={classes.root} direction="column">
+      <Backdrop className={classes.backdrop} open={isLoading}>
```
mark-prins, line 81: nice!

unified-codes/apps/web/src/components/templates/Details.tsx
```diff
@@ -2,41 +2,107 @@ import * as React from 'react';
...
+export const DetailsComponent: Details = ({ classes, entity, isLoading, getEntity }) => {
   const { code } = useParams<IDetailsParameters>();
 
   React.useEffect(() => {
     getEntity(code);
   }, []);
 
-  // TODO: Create UI components and format this properly
-  return <div>{JSON.stringify(entity)}</div>;
+  let productSubCategories: IEntity[] = [];
```
mark-prins, line 54: these can be `const` as they aren't reassigned. The array has items pushed into it, that's all.

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailCategoryItem.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+
+import { IEntity } from '@unified-codes/data';
+
+export interface EntityDetailCategoryItemProps {
+  entity: IEntity;
+}
+
+export type EntityDetailCategoryItem = React.FunctionComponent<EntityDetailCategoryItemProps>;
```
mark-prins, line 9: almost getting too long as a name.. would `EntityCategory` be descriptive enough?

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailList.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  return (
+    <Grid direction="column">
+      <Typography variant={headerTypography}>Forms</Typography>
+      <ul>
+        {productSubCategories?.map((category) => (
+          <EntityDetailCategoryItem entity={category} />
+        ))}
+      </ul>
+      <Typography variant={headerTypography}>Properties</Typography>
```
mark-prins, line 40: I wonder about not showing a section if there are no elements in it?
Either that or showing 'none' under the heading?

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailList.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <ul>
+        {entityProperties?.map((item) => (
+          <li key={item.type}>
+            {item.type}: {item.value}
+          </li>
+        ))}
+      </ul>
+      <Typography variant={headerTypography}>External Links</Typography>
```
mark-prins, line 48: A refactoring suggestion if the hiding of sections is a useful idea:
```
const DetailSection = ({
  headerTypography,
  items,
  title,
}: {
  headerTypography: any;
  items: Array<JSX.Element>;
  title: string;
}) => (
  <Grid item>
    <Typography variant={headerTypography}>{title}</Typography>
    <ul>{items}</ul>
  </Grid>
);

...

const properties = entityProperties?.map((item) => (
    <li key={item.type}>
      {item.type}: {item.value}
    </li>
  ));

...

<DetailSection headerTypography={headerTypography} items={properties} title="Properties" />
```

etc. same pattern for the other two. Just a thought.
const detailSection

mark-prins (approved):
Looks good!
I have suggested a thing - not a show stopper at all. And we can tweak design later, looks ok for now!

unified-codes/apps/web/src/components/templates/Details.tsx
katherine-sussol: Oh - thanks, that's a good point. I must have been thinking of a different kind of array at the time. Updated! 63e5bdf

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailCategoryItem.tsx
katherine-sussol: I do agree it's getting too long 😞  - another case of struggling with names. I was thinking `Detail` would help denote it as specific to `EntityDetails` and not related to the `EntityTable` used in the explorer. Would `EntityDetailCategory` work? 🤔  Still a little long

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailList.tsx
katherine-sussol: I like the idea of showing 'none' under the header (shows there is potentially some stuff but not for that particular product).

unified-codes/apps/web/src/components/templates/Details.tsx
```diff
@@ -2,41 +2,107 @@ import * as React from 'react';
...
 
 export type Details = React.FunctionComponent<DetailsProps>;
 
-export const DetailsComponent: Details = ({ entity, getEntity }) => {
+export const DetailsComponent: Details = ({ classes, entity, isLoading, getEntity }) => {
   const { code } = useParams<IDetailsParameters>();
 
   React.useEffect(() => {
     getEntity(code);
   }, []);
```
wlthomson, line 52: Not very familiar with `useParams`, so apologies if I'm a bit confused by this logic!

Is the `code` coming from the navigation props and being used to then fetch the `entity`, which is then passed back through as a prop?

If you're OK with me refactoring as I go, then I think this stuff should be put into the redux store, thus avoiding the rerenders and stuff. Rather than navigating, would suggest just having this as a `page` component inside a template, and just updating the store with the selected `IEntity`, e.g. `selectedEntity` (on click callback should provide `index` or some other info which can identify this, I think). From that, can just use selectors in `mapStateToProps` to extract all the metadata that is being extracted in the component. 

If we do that, we essentially just need a presentational component which does the rendering bits (not sure whether it's best in the library or web app, though).

Does that make sense?

unified-codes/libs/data/src/lib/types/Entity.ts
```diff
@@ -4,20 +4,23 @@ export interface IEntity {
   code: string;
   description: string;
   type: string;
-  properties?: Property[];
+  has_property?: Property[];
+  has_child?: IEntity[];
```
wlthomson, line 8: I think dgraph syntax of relations might be causing confusion! think this should be `properties` and `children` (although not too sure about having a recursive interface, but think it's OK!). If you need queries, can add them to the `Entity` class, e.g.

```
hasChild(child: IEntity) { 
  return this.children.map((child: IEntity) => child.code).includes(child.code); 
}
```

or something like that (or could also continue with existing pattern with something like `matchesChild`)?

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailCategoryItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const EntityDetailCategoryItem: EntityDetailCategoryItem = ({ entity }) => {
+  return (
+    <li key={entity.description}>
+      {entity.description} {`(code:${entity.code})`}
+      <ul>
+        {entity.has_child?.map((child) => (
+          <EntityDetailCategoryItem entity={child} />
```
wlthomson: I think this is including too much logic for a presentational component? I think this is stuff that should be controlled by the parent. E.g. unless I'm mistaken, the parent passes an entity but doesn't have control over depth of recursion?

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailCategoryItem.tsx
wlthomson: I think when the names for components start getting long it's usually an indication of thinking in OOP inheritance instead of composition (as opposed to big long interface names, which are quite common, I think because of OOP's use of inheritance)! I think this component may be a bit too specific, could probably handle the list logic in the parent (possibly using a generic presentational component to display it)?

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailList.tsx
```diff
@@ -0,0 +1,60 @@
...
+export interface EntityDetailListProps {
+  headerVariant?: TypographyVariant;
+  productSubCategories?: IEntity[];
+  entityProperties?: Property[];
+  externalLinks?: IExternalLink[];
+}
+
+export type EntityDetailList = React.FunctionComponent<EntityDetailListProps>;
+
+export const EntityDetailList: EntityDetailList = ({
```
wlthomson, line 24: I think this component is probably useful, but again must be annoying and say it contains logic 😢. Ideally library components should just take props and render them, not derive data and stuff.

wlthomson (approved):
Approving to keep things moving!

My only concern is that the new stuff being done is (understandably) following the existing patterns and not the reworked redux-driven way. Only issue is that this means that as new stuff goes in, the refactor branch falls behind by that much. Shouldn't be a problem, but I might need some help at some stage with the other branch to catch up enough that we can do a merge! 

go

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailCategoryItem.tsx
wlthomson: Also, recursion in react is a pretty dangerous path to go down because of how the render cycle works and stuff (generally would avoid components rendering themselves unless you really know what you're doing with the render cycle and stuff... which I don't!).

unified-codes/apps/web/src/components/templates/Details.tsx
katherine-sussol: Yeah - it's a param fetched from the react-router (mapped to code via `/:code` in App.tsx). The reason I went with this was so we had the ability to direct link to a specific product page - e.g. I could give you http://localhost:4200/details/c7750265 and you wouldn't need to do a search using the Explorer. 

I see what you're saying though, in that when the code is set (either via react-router or from `selectedEntity` event), it could go into the store rather than being read directly from this template? That could work too - I think.

unified-codes/libs/data/src/lib/types/Entity.ts
katherine-sussol: Yeah, I do agree children and properties sound nicer. I think the GraphQL API (the resolver anyway) needs some tweaking before we can use those names though - I was getting no results when using `children` and `properties` 😞

unified-codes/libs/ui/src/lib/components/entity/molecules/EntityDetailCategoryItem.tsx
katherine-sussol: Hm - I see what you're saying. Do you think it's worth specifying a maxDepth parameter on the parent? With our current dataset, it shouldn't be a problem, and the saga is currently only fetching a max of 3 children - but that could change in the future.

wlthomson: Generally, I think library components which directly wrap data structures are probably to be avoided as they can be defined by a parent component (as they are often dynamic anyway...). In this instance the component exists to basically map an `entity` to HTML tags (should prob use library components for these instead, if possible). 

I think comes back to what I was saying re. composition v. OOP. OOP encourages more specific instantiations to handle specific cases, whereas composition encourages composing generic components together. Generally library components should just be presentational, i.e. think of them as functions whose purpose is just to directly display props they are given (as opposed to transformational functions which map/reduce inputs to create new data structures).

Does that make sense? I wouldn't worry about it now, I'll catch up soon with the refactor anyway!

unified-codes/apps/web/src/components/templates/Details.tsx
wlthomson: I think can still use navigation, but I think the props should be given to the component directly from the store. I think it's a bit of a code smell to be using `useEffect` for initialising the component (exception can be for data which has to be fetched remotely after the component is rendered).

I think ideally, the component should take the `entity` directly (again comes back to avoiding as much logic in the component as possible). The easiest way to do this is to just have all the `EntityDetails` component state in the redux store (like what I'm doing with explorer). That way when you're designing the components, you can just design them with the props you want, rather than having to tightly couple them to navigation (which `useParams` does, as it makes the component unusable without navigation 😢 ).

Does that make sense? Again, wouldn't worry for now, I'll catch up soon!

katherine-sussol:
Thanks for the feedback @wlthomson - appreciate it! Do see your points about the target composition/structure, just wondering now is how we manage these changes with the refactor, would you prefer to leave this as a WIP PR or shall I merge it for it to be adjusted later? Do think it's probably good to avoid having too many in-progress PRs but don't want to disrupt anything you're doing 😅

katherine-sussol:
Update: Will merge - to be tidied up via separate issues (and a disclaimer that I am not personally attached to them and welcome input 📝 , killing 🪓 ,  and mutations 🧟.).

PR #215, opened 2020-10-05
#212 migrate entity details saga

### LINKED ISSUE #212: Migrate entity details saga to updated app structure
## Expected Behaviour

Same entity fetching functionality as the current master branch - but in the fancy new structure.

katherine-sussol:
Fixes #212.

## Description

Branched off #213.

Re-adds saga to fetch details in new code structure. NB: Details page currently loses the entity on refresh as it expects the entity to be set from the explorer - probably something to fix when re-visiting navigation..?

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.

### DIFF
--- a/unified-codes/apps/web/src/actions/DetailActions.ts
+++ b/unified-codes/apps/web/src/actions/DetailActions.ts
@@ -3,58 +3,46 @@ import { Action } from 'redux';
 
 export const DETAIL_ACTIONS = {
   UPDATE_ENTITY: 'detail/updateEntity',
-  FETCH_ENTITY: 'detail/fetchEntity',
-  FETCH_ENTITY_SUCCESS: 'detail/fetchEntitySuccess',
-  FETCH_ENTITY_FAILURE: 'detail/fetchEntityFailure',
+  UPDATE_ENTITY_SUCCESS: 'detail/updateEntitySuccess',
+  UPDATE_ENTITY_FAILURE: 'detail/updateEntityFailure',
 };
 
 export interface IDetailUpdateEntityAction extends Action<string> {
   entity: IEntity;
 }
 
-export interface IDetailFetchEntityAction extends Action<string> {
-  code: string;
-}
-
-export interface IDetailFetchSuccessAction extends Action<string> {
+export interface IDetailUpdateSuccessAction extends Action<string> {
   entity: IEntity;
 }
 
-export interface IDetailFetchFailureAction extends Action<string> {
+export interface IDetailUpdateFailureAction extends Action<string> {
   error: Error;
 }
 
 export type IDetailAction =
   | IDetailUpdateEntityAction
-  | IDetailFetchEntityAction
-  | IDetailFetchSuccessAction
-  | IDetailFetchFailureAction;
+  | IDetailUpdateSuccessAction
+  | IDetailUpdateFailureAction;
 
 const updateEntity = (entity: IEntity) => ({
   type: DETAIL_ACTIONS.UPDATE_ENTITY,
   entity,
 });
 
-const fetchEntity = (code: string) => ({
-  type: DETAIL_ACTIONS.FETCH_ENTITY,
-  code,
-});
-
-const fetchEntitySuccess = (entity: IEntity) => ({
-  type: DETAIL_ACTIONS.FETCH_ENTITY_SUCCESS,
+const updateEntitySuccess = (entity: IEntity) => ({
+  type: DETAIL_ACTIONS.UPDATE_ENTITY_SUCCESS,
   entity,
 });
 
-const fetchEntityFailure = (error: Error) => ({
-  type: DETAIL_ACTIONS.FETCH_ENTITY_FAILURE,
+const updateEntityFailure = (error: Error) => ({
+  type: DETAIL_ACTIONS.UPDATE_ENTITY_FAILURE,
   error,
 });
 
 export const DetailActions = {
   updateEntity,
-  fetchEntity,
-  fetchEntitySuccess,
-  fetchEntityFailure,
+  updateEntitySuccess,
+  updateEntityFailure,
 };
 
 export default DetailActions;
--- a/unified-codes/apps/web/src/actions/DetailActions.ts
+++ b/unified-codes/apps/web/src/actions/DetailActions.ts
@@ -3,21 +3,58 @@ import { Action } from 'redux';
 
 export const DETAIL_ACTIONS = {
   UPDATE_ENTITY: 'detail/updateEntity',
+  FETCH_ENTITY: 'detail/fetchEntity',
+  FETCH_ENTITY_SUCCESS: 'detail/fetchEntitySuccess',
+  FETCH_ENTITY_FAILURE: 'detail/fetchEntityFailure',
 };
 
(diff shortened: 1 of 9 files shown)


unified-codes/apps/web/src/actions/DetailActions.ts
```diff
@@ -3,21 +3,58 @@ import { Action } from 'redux';
...
+  code: string;
+}
+
+export interface IDetailFetchSuccessAction extends Action<string> {
+  entity: IEntity;
+}
+
+export interface IDetailFetchFailureAction extends Action<string> {
+  error: Error;
+}
```
wlthomson, line 25: I think the `FetchEntityAction` is redundant (as every time the entity is updated a fetch should probably occur?). Can probably go without and just rename the success and failures as `IDetailUpdateSuccessAction` etc.? If we need a separate action can always add it when needed 🤷 .

unified-codes/apps/web/src/components/explorer/ExplorerTableRows.tsx
```diff
@@ -11,29 +11,36 @@ import { IState } from '../../types';
...
-}
+  return <EntityTableRows {...props} onSelect={onSelect} />;
+};
 
 const mapDispatchToProps = (dispatch: React.Dispatch<IExplorerAction>) => {
-    const onSelect = (entity: IEntity) => dispatch(DetailActions.updateEntity(entity));
-    return { onSelect };
+  const onSelect = (entity: IEntity) => {
+    dispatch(DetailActions.updateEntity(entity));
+    dispatch(DetailActions.fetchEntity(entity.code));
```
wlthomson, line 28: If follow above, no need for the `fetchEntity` here (woo)!

unified-codes/apps/web/src/reducers/DetailReducer.ts
```diff
@@ -17,6 +17,17 @@ export const DetailReducer = (state: IDetailState = initialState(), action: IDet
             const { entity } = action as IDetailUpdateEntityAction;
             return { ...state, entity }
         }
+        case DETAIL_ACTIONS.FETCH_ENTITY: {
+          return { ...state, loading: true }
+        }
+        case DETAIL_ACTIONS.FETCH_ENTITY_SUCCESS: {
+          const { entity } = action as IDetailFetchSuccessAction;
+          return { ...state, entity, error: initialState().error, loading: false };
```
wlthomson, line 25:
```suggestion
        case DETAIL_ACTIONS.FETCH_ENTITY_SUCCESS: {
          const { entity } = action as IDetailFetchSuccessAction;
          const { error } = initialState;
          return { ...state, entity, error, loading: false };
```

unified-codes/apps/web/src/reducers/DetailReducer.ts
```diff
@@ -1,10 +1,10 @@
-import { DETAIL_ACTIONS, IDetailAction, IDetailUpdateEntityAction } from '../actions';
+import { DETAIL_ACTIONS, IDetailAction, IDetailUpdateEntityAction, IDetailFetchEntityAction, IDetailFetchSuccessAction, IDetailFetchFailureAction } from '../actions';
 import { IDetailState } from '../types';
 
 const initialState = (): IDetailState => {
```
wlthomson, line 4: Whoops, this was me! 

Recommend changing this to an object instead of function (see `ExplorerReducer` for e.g.).

unified-codes/apps/web/src/sagas/DetailSaga.ts
```diff
@@ -0,0 +1,112 @@
...
+    has_child {
+      code
+      description
+      type
+      # form
+      has_child {
+        code
+        description
+        type
+        # unit_of_use
```
wlthomson, line 59: I think discussed this with @mark-prins. Not necessarily a problem (for now), but I think this may be complicated as this only works if we assume we will always be fetching `Drug` entities (as other entities may not have forms etc.)?

unified-codes/apps/web/src/sagas/DetailSaga.ts
```diff
@@ -0,0 +1,112 @@
...
+  const { data } = json;
+  const { entity } = data;
+  return entity;
+};
+
+function* fetchDetails(action: IDetailFetchEntityAction) {
+  yield put(AlertActions.raiseAlert(alertFetch));
+  try {
+    const url = `${process.env.NX_DATA_SERVICE_URL}:${process.env.NX_DATA_SERVICE_PORT}/${process.env.NX_DATA_SERVICE_GRAPHQL}`;
+    const entity: IEntity = yield call(getEntity, url, action.code);
```
wlthomson, line 94:
```suggestion
    const { code } = action;
    const entity: IEntity = yield call(getEntity, url, code);
```

wlthomson (approved):
Few nits, nothing major.

thumbsup

unified-codes/apps/web/src/sagas/DetailSaga.ts
katherine-sussol: Ah yes, bit tidier :) thanks I will use the suggestion

unified-codes/apps/web/src/sagas/DetailSaga.ts
katherine-sussol: Currently if there are no forms it will just fetch whatever child happens to be there (so yeah the comment is probably a little misleading - oops). E.g. for `Other` this will just be `unit_of_use` with the current test data. I did a bit of testing also, if there are no children it just returns a null object for the child. 

Agree this is not very nice on the client-side as we start having to read types and things to parse the data since everything is an `Entity`. Entities for everyone 🥳

unified-codes/apps/web/src/reducers/DetailReducer.ts
katherine-sussol: Haha I didn't even notice. I'll update it

unified-codes/apps/web/src/actions/DetailActions.ts
katherine-sussol: Yeah I wasn't 100% about having a separate action or not. I think you're right though - since we don't currently have a reason to have them separate.

katherine-sussol: Updated in d5c829867

unified-codes/apps/web/src/sagas/DetailSaga.ts
mark-prins: I would `recurse` to a depth of 4, as that would pick up the properties on the unit of use as well.

mark-prins (approved):
Looks good! thanks @katherine-sussol

unified-codes/apps/web/src/sagas/DetailSaga.ts
katherine-sussol: Hmm, I'm not sure how to do this (`recurse`) in pure GraphQL (though it's possible in DgraphQL). Would adding a `has_property` field to each of the children provide a sufficient level of detail? Or do we also want children of `unit_of_use`?

mark-prins: oh, right. ignore the comment then, sorry! merge ahoy!

PR #237, opened 2020-10-15
#235 connect pagination

### LINKED ISSUE #235: Connect pagination
The pagination actions ( rows per page & current page ) aren't doing anything. Just need to connect up the actions

## Expected Behaviour
To be able to view next/previous page of results and change the rows per page.

mark-prins:
Fixes #235 

## Description
Connects pagination controls - have added the ability to pass in parameters when updating entities; this allows updating of params for the search, without needing to chain dispatches or wait for the store to update before fetching.

Other suggestions?

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


unified-codes/apps/web/src/actions/ExplorerActions.ts
```diff
@@ -1,247 +1,252 @@
...
 }
 
 export interface IExplorerTableUpdateFilterByAction extends Action<string> {
-    filterBy: string;
+  filterBy: string;
 }
 
-export interface IExplorerTableFetchEntitiesAction extends Action<string> { };
+export interface IExplorerTableFetchEntitiesAction extends Action<string> {
+  parameters?: IExplorerParameters;
```
wlthomson, line 34: I think this is unnecessary? I think either the saga should use the selectors, or we should pass them via the action.  Since the store is the source of truth, I think it doesn't really matter which way we go, but I think we should choose one?

unified-codes/apps/web/src/sagas/ExplorerSaga.ts
```diff
@@ -48,33 +54,16 @@ const getEntities = async (url: string, query: ExplorerQuery): Promise<IEntity[]
...
-        description,
-        types,
-        orderBy,
-        orderDesc,
-        rowsPerPage,
-        page,
-      };
-
+      const parameters: IExplorerParameters =
+        action?.parameters || (yield select(ExplorerSelectors.selectParameters));
```
wlthomson, line 65: As above 😄 .

unified-codes/apps/web/src/selectors/ExplorerSelectors.ts
```diff
@@ -79,6 +80,39 @@ const selectRowsPerPage = createSelector(
 
 const selectPage = createSelector(selectTable, (table: IExplorerTableState): number => table?.page);
 
+const selectTypes = createSelector(
+  selectToggleBar,
+  (toggleBar: IExplorerToggleBarState): EEntityType[] =>
+    Object.keys(toggleBar).filter((type: EEntityType) => toggleBar[type]) as EEntityType[]
+);
```
wlthomson, line 87: I think this is going against the design decision of having the types explicit in the store?

unified-codes/apps/web/src/components/containers/explorer/ExplorerSearchBar.tsx
```diff
@@ -10,31 +10,40 @@ import { IState } from '../../../types';
...
-        dispatch(ExplorerActions.updateEntities());
+  const onChange = (input: string) => dispatch(ExplorerActions.updateInput(input));
+
+  const onClear = () =>
+    batch(() => {
+      dispatch(ExplorerActions.resetPage());
+      dispatch(ExplorerActions.resetInput());
+      dispatch(ExplorerActions.resetFilterBy());
+      dispatch(ExplorerActions.updateEntities());
     });
```
wlthomson, line 26: Even though this is in a `batch`, it feels like `resetPage` makes more sense at the end rather than the beginning?

wlthomson (approved):
A few minors, but mostly LGTM!

unified-codes/apps/web/src/actions/ExplorerActions.ts
wlthomson: Oops, just read the description: 

```
this allows updating of params for the search, without needing to chain dispatches or wait for the store to update before fetching.
```

I think we want to maintain the "state machine" logic. So I think chaining dispatches is probably the best way to do it... otherwise it can end up introducing complexities and tricky bugs. I don't think there;s a performance need to do this, so I would suggest keeping it simple? Thoughts?

wlthomson:
Probably my mistake for not having the repo formatted, but I think we should try to avoid formatting changes in PRs which also introduce other changes! Makes it really hard to see what's been changed (I know I've probably done this and a lot worse)!

unified-codes/apps/web/src/components/containers/explorer/ExplorerTablePagination.tsx
```diff
@@ -1,52 +1,98 @@
...
 
 const mapDispatchToProps = (dispatch: React.Dispatch<IExplorerAction>) => {
-    const onChangePage = (_: React.MouseEvent<HTMLButtonElement> | null, page: number) => dispatch(ExplorerActions.updatePage(page));
-    const onChangeRowsPerPage = (event: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) => dispatch(ExplorerActions.updateRowsPerPage(+event.target.value));
-    
-    return { onChangePage, onChangeRowsPerPage };
+  const onChangePage = (page: number, parameters: IExplorerParameters) =>
+    batch(() => {
+      dispatch(ExplorerActions.updatePage(page));
+      dispatch(ExplorerActions.updateEntities({ ...parameters, page }));
```
wlthomson, line 70: As above, I think best to avoid this sort of thing. I don't think it will churn out more performance, and it introduces multiple ways of doing the same thing, which I think is possibly likely to introduce confusion/bugs?

wlthomson: Also, ideally we want the source-of-truth to always be "up-to-date". Giving the saga access to newer data before the store is updated feels wrong to me (just me?).

wlthomson: Also, does this behave differently when the parameters are removed? I think `batch` only effects render cycles, I think the dispatches are still performed sequentially?

katherine-sussol (approved):
I think I agree with the formatting being a bit distracting but all good otherwise 😸

unified-codes/apps/web/src/components/containers/explorer/ExplorerTablePagination.tsx
mark-prins: true, re the `batch` - this only affects render cycles and isn't helpful here.
Dispatches are themselves synchronous, but I understood that the store isn't necessarily synchronously updated. I think though, that I might be confusing React behaviour - react will asynchronously batch updates to components, so the state which a component sees isn't updated directly after an action is dispatched. 

I'll rework and see if it hangs together. I'm concerned that if it doesn't run synchronously then that particular bug will be difficult to discover.

mark-prins:
Ok - reverted the worst parts and updated after formatting master. Only 4 files changed now 😃

Issue #171, opened 2020-09-08
Data service schema mandatory fields


wlthomson:
While reviewing #169, noticed a possible discrepancy with some of the predicates for the `Entity` type.

## Expected Behaviour

`Entity` type should have mandatory `code` (and `type`?), and optional description.

```
type Entity {
  code: String!
  description: String
  has_property: [Property]
  has_child: [Entity]
  type: String
}
```

## Current Behaviour

`Entity` type has mandatory `type` and `description`, and optional `code`:

```
type Entity {
  type: String!
  code: String
  description: String!
  has_property: [Property]
  has_child: [Entity]
}
```

## Possible Solution

Update the schema? 

Note: if others are in agreement, I think it would be nice to also have a simple convention for schemas, whereby predicates are listed alphabetically (might just simply things)? If so, could possibly piggyback that into this.

🐷  (piggyback request): can we remove the `code` search as part of this please? ( refer to #172 )
🐷  (piggyback request): make orderBy and offset / first mandatory, or perhaps default those that make sense to default. ( reference: https://github.com/openmsupply/unified-codes/pull/178#discussion_r488305353 )
🐷  (piggyback request): Confirm DetailEntityListItem key is unique https://github.com/openmsupply/unified-codes/pull/249#discussion_r508075613

wlthomson:
@katherine-sussol @mark-prins Not sure if this is me missing something, so marking this as a discussion/question rather than bug for now!

mark-prins:
Yes! alphabetical makes me happy. Agree with the mandatoryness too. Noticed that and thought to change but haven't yet.
![](https://media1.giphy.com/media/13G7hmmFr9yuxG/giphy.gif)

katherine-sussol:
Agree with both alphabeticalling and mandatorying! I think I tried to make code mandatory once but ran into validation issues as the current migrated data has some entities without codes (and also wasn't sure if it was truly mandatory and didn't want to dig too deep) - in particular, things like form entities, should it be possible to create lots of { type: form, description: injection } nodes with different codes (currently they're different nodes but with no code)? Or should all injections inherit from the same node? Graph wise I think it makes sense to refer to the same node? 

Either way (and sorry if I'm being confusing) I think some data massaging may be required 💆

mark-prins:
@katherine-sussol oh, yes - blank `code` would be a problem!
[...]

PR #249, opened 2020-10-18
#246 missing prop key and grid direction

### LINKED ISSUE #246: Missing prop key and grid direction on details page
## Expected Behaviour

No console warnings 

## Current Behaviour

1. Screen Shot 2020-10-16 at 5 08 01 PM
2. Screen Shot 2020-10-16 at 5 09 04 PM

## Possible Solution

- Add the missing props. Somewhere

katherine-sussol:
Fixes #246.

## Description

- Minor fixes to get rid of the pesky red console messages on the details page.
- Removed an extra `u` character that snuck its way into the `data-service `resolver

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
```diff
@@ -63,7 +63,7 @@ const DetailEntityListItem: DetailEntityListItem = ({ description, childEntities
     if (!childCount) return null;
     const childItems = childEntities?.map((child: IEntity) => {
       const { description, children } = child;
-      return <DetailEntityListItem description={description} childEntities={children} />;
+      return <DetailEntityListItem description={description} childEntities={children} key={child.description} />;
```
wlthomson, line 66: Suggest using `code` for the key, since it's unique?

wlthomson:
```suggestion
      const { code, description, children } = child;
      return <DetailEntityListItem key={code} description={description} childEntities={children} />;
```

unified-codes/apps/web/src/components/containers/detail/DetailPropertyListItem.tsx
```diff
@@ -64,7 +64,7 @@ const DetailPropertyListItem: DetailPropertyListItem = ({ description, propertie
     const childProperties = properties?.map((property: IProperty) => {
       const { type, value, properties } = property;
       const description = `${type}: ${value}`;
-      return <DetailPropertyListItem description={description} properties={properties} />;
+      return <DetailPropertyListItem key={description} description={description} properties={properties} />;
```
wlthomson, line 67: This one is a bit trickier, but it feels a bit yucky to use a UI string for logic? Since properties won't be duplicated for the same parent, suggest just using `key={type]`.

wlthomson:
```suggestion
      return <DetailPropertyListItem key={type} description={description} properties={properties} />;
```

unified-codes/apps/web/src/components/layout/DetailLayout.tsx
```diff
@@ -21,7 +21,7 @@ export const DetailLayout: DetailLayout = ({ classes, attributeList, childList,
     <Grid item className={classes?.attributeListContainer}>
       {attributeList}
     </Grid>
-    <Grid item direction="row" className={classes?.childListContainer}>
+    <Grid item className={classes?.childListContainer}>
```
wlthomson, line 24: Does this effect the UI layout? I was thinking just to add a `container` prop?

wlthomson:
```suggestion
    <Grid container item className={classes?.childListContainer}>
```

unified-codes/apps/data-service/src/app/resolvers.ts
```diff
@@ -80,7 +80,6 @@ export class EntityResolver {
     const order = `order${orderBy.descending ? 'desc' : 'asc'}: ${orderBy.field}`;
     const query = queries.entities(type, order, offset, first, description);
     const response = await dgraph.postQuery(query);
-    u;
```
wlthomson, line 83: Oops!

wlthomson (changes requested):

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
katherine-sussol: I agree with you that using description is icky but unfortunately, code doesn't work at the moment as there are currently many intermediary nodes that are `null` which results in duplicate key errors. 

I think we either adjust this to use code with #171 or expose the underlying `uid` and use that (but I personally feel this is even worse as the UI shouldn't know about Dgraph uids).

unified-codes/apps/web/src/components/containers/detail/DetailPropertyListItem.tsx
katherine-sussol: Cool yeah, type is probably better - I don't think we have any duplicates at the moment (can revisit if it does happen) so should work nicely.

unified-codes/apps/web/src/components/layout/DetailLayout.tsx
katherine-sussol: `container` prop works just as well visually - happy to change it.
What about the `attributeList` and `propertyList` containers, should we convert those too for consistency?

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
wlthomson: Agree with first option + that no nodes should have `null` codes.

unified-codes/apps/web/src/components/layout/DetailLayout.tsx
wlthomson: Yup, if it doesn't change the UI appearance then go for it!

mark-prins: @wlthomson according to material-UI docs: "_If true, the component will have the flex item behavior. You should be wrapping items with a container._"

how then, can you add both `item` and `container` to a `Grid`? which one is it?

katherine-sussol: Coool 😎  - 2e977f145

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
katherine-sussol: Sounds good to me - added a note to the issue description so we don't forget (issue now has 3 little 🐷 )

mark-prins (commented):
Looks good!
Would be nice to have `code` as the `key` and to use the `container` prop to indicate row direction, as noted.

perhaps re-import using the current data.json?

I've queried and have no entities with no code :)

```
{
    query (func: eq(type,"unit_of_use"), orderasc: description) @filter(not has(code)) {	
      code
      description
      type
      uid
      value
      has_property
    }
```

unified-codes/apps/web/src/components/layout/DetailLayout.tsx
katherine-sussol: @mark-prins sorry I think my UI hadn't refreshed and I didn't see your comment when making the commit. I decided to just use a container 😸

katherine-sussol:
Hmm I'm still getting `null` codes with what I think is the latest json file - is there are more recent one 😮 ? (have checked `data.json` in `tools\scripts\dgraph\data` and also `data.json` in the universal-drug-codes-server repo):
```
"entity": {
      "code": "4231441c",
      "description": "Amoxicillin",
      "type": "drug",
      "children": [
        {
          "code": null,
          "description": "oral",
          "type": "form_category",
          "children": [
            {
              "code": null,
              "description": "capsule",
              "type": "form",
              "children": []
[...]
```

mark-prins:
ah. and oh. and yeah. the forms and form_categories don't have codes.
should we add them, even though those things don't need to have codes?
I'm thinking 'might as well' as that makes the data consistent, and there's little chance of running out of codes.

katherine-sussol:
Yeah that sounds good - was thinking it would be something done as part of #171 as that included making code mandatory.
Think what I will do is just make the change to use code here (so it's doing the right thing from this PR perspective) and then pick up #171 to fix the duplicate key error

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
katherine-sussol: ed1c3c5

mark-prins (approved):
LGTM

PR #251, opened 2020-10-19
#250 add data source classes

wlthomson:
Fixes #250.

## Description

A minor tidy up to the back-end interactions code. Also added some comments and helper types to make make the code a bit more self-documenting.

### Documentation:

Tick below if one of the following applies:

1. Documentation has been added to cover changes introduced by this PR.
2. This PR requires no changes to documentation.

- [ ] Documentation OK.

### Tests:

Tick below if one of the following applies:

1. Tests have been added to cover changes introduced by this PR.
2. This PR requires no new tests.

- [ ] Tests OK.

PR #255, opened 2020-10-19
#252 add interactions default resolver

### LINKED ISSUE #252: Expose interactions using default resolver
Interactions are resolved via the RxNav API, which returns relatively big payloads and which doesn't support batch queries. This is problematic as it causes query delays and may cause users to misuse the RxNav API and risk the UC server being throttled or blacklisted.

The current solution is that interactions are only exposed via the `entity` query path. For now, this is handled by only exposing interaction as a field via the root `entity` query. 

## Expected Behaviour

Same as now.

## Current Behaviour

Interactions are exposed via a field resolver which contains a conditional which checks the query path for the root query. 

## Possible Solution

Migrate the field resolver to a function returned on the `entity` query (see Apollo docs on [default resolvers]( https://www.apollographql.com/docs/apollo-server/data/resolvers/#default-resolvers)).

wlthomson:
Fixes #252.

Branched off #251.

## Description

Migrates data source + resolver logic to data source classes.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.

### DIFF
--- a/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
@@ -1,6 +1,12 @@
 import { RESTDataSource } from 'apollo-datasource-rest';
 
-import { EntityCollection, IEntity, IEntityCollection, EEntityField, EEntityType, Entity } from '@unified-codes/data';
+import {
+  EEntityField,
+  EEntityType,
+  EntityCollection,
+  IEntity,
+  IEntityCollection,
+} from '@unified-codes/data';
 
 export class DgraphDataSource extends RESTDataSource {
   private static headers: { [key: string]: string } = {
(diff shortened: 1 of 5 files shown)


unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
```diff
@@ -75,9 +81,12 @@ export class DgraphDataSource extends RESTDataSource {
...
     );
 
-    const { counters, query } = data ?? {};
-    const entities: IEntity[] = query ?? [];
-    const [{ total: totalCount }] = counters ?? [];
+    const { counters: counterData, query: entityData } = data ?? {};
+    const [totalCount] = counterData?.total ?? [];
+    // Overwrite interactions to prevent large query delays.
+    const entities: IEntity[] =
+      entityData?.map((entity: IEntity) => ({ ...entity, interactions: [] })) ?? [];
```
katherine-sussol, line 88: Oh nice 😄

katherine-sussol (approved):
Sweet 🍬  looks good to me.

Thanks for another 🐈  artifact tidy up 🧹 👍

unified-codes/libs/data/src/lib/types/Entity.ts
```diff
@@ -56,6 +58,16 @@ export class Entity implements IEntity {
     return this._properties;
   }
 
+  getChild(code: string) {
+    const [entity] = this._children?.filter((child: IEntity) => child.code === code);
```
katherine-sussol, line 62: Oh sorry - minor thing I just noticed from TS on this line (but it does still run fine):
`Type 'Entity[] | undefined' must have a '[Symbol.iterator]()' method that returns an iterator.`

and a similar one for getProperty

wlthomson: Hmmm, c718a7985c9f22442b3f9637d0e0f64caced16f3 should fix that!

PR #258, opened 2020-10-22
#171 data service schema mandatory fields

### LINKED ISSUE #171: Data service schema mandatory fields

katherine-sussol:
Fixes #171 

## Description

Thanks @mark-prins for the code generation for the `form` and `form_category` nodes - I stole them and updated `import.json` with the latest changes so `data-service` works without needing to pull from the other repo.

- `code`, `type`, `EntitySortInput` made mandatory
- Fields alphabetised within type definitions in schema
- This (just the updated `import.json` actually) - fixes the duplicate code errors that show up after #249

Realise our schema is subject to change soon anyway, but better to keep the repo in a healthy state in case of any ad hoc demo requests 😄 

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


wlthomson (approved):
LGTM.

One concern is I'm not so sure about the sorting being mandatory. I think it was my feedback initially which prompted sorting to be split out into a dedicated input, but I think I made an error there in hindsight! IMO the `entities` should be fetchable without a mandatory sorting (just use default alphabetic sort by description or something)?

katherine-sussol:
Good point - I agree with you, a bit of a weird option to have mandatory. Have modified to use default behaviour instead when not specified with 5925a53ed (left as a dedicated input but an optional one)

PR #275, opened 2020-11-02
implement match filter element

### LINKED ISSUE #259: Narrow entity query search
When searching for a matching item using mSupply, the current query which searches for 'contains' is too broad.
Add the ability to specify 'contains' or 'fuzzy' and default to 'starts with'

## Expected Behaviour
Searching for `Midazolam` should return products starting with `Mid` and not the current list, in which the top 25 doesn't include anything starting with 'M'

mark-prins:
Fixes #259 

## Description
Implement a `match` parameter on the `filter` property.
This optional value can be `exact | begins (default) | contains`

Can be tested using postman

```
{ "query": "{ entities(filter: { code: \"\" description: \"Dia\" type: \"[drug]\" match: \"contains\" }  offset: 0 first: 25) { data { code description type uid }, totalLength, }}"}
```

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.

### DIFF
--- a/unified-codes/apps/data-service/src/app/schema.ts
+++ b/unified-codes/apps/data-service/src/app/schema.ts
@@ -1,4 +1,4 @@
-import { Field, ID, InputType, Int, ObjectType, registerEnumType } from 'type-graphql';
+import { Field, ID, InputType, Int, ObjectType /*, registerEnumType */ } from 'type-graphql';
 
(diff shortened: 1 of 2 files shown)


katherine-sussol (approved):
👀 s good to me

PR #277, opened 2020-11-03
#276 return properties of parent

### LINKED ISSUE #276: Return properties of parent
For entities at the `unit_of_use` level, some properties should be sourced from the parent not from the current node.
These are
- WHO EML categorisation
- RxCUID when used for interactions

## Expected Behaviour
When searching in mSupply for the product `Amoxicillin 250mg capsule` the property `who_eml` should come from the parent product `Amoxicillin` 

When searching for `Ampicillin 125mg/5mL oral liquid` the `code_rxnav` should be sourced from the parent, returning `733` - currently no interactions are returned.

Suggest prioritising property values: providing the immediate properties if specified, and reverting to parent properties if not. Only for the two properties specified though.

## Current Behaviour
Currently the WHO EML value is replicated at the `unit_of_use` level which avoids the problem. with the data changes this will become a problem and the EML property won't be returned for an entity.

mark-prins:
Fixes #276 

## Description
Updates an entities props - taking product props if appropriate
Used a product rx id for interactions

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.

### DIFF
--- a/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
@@ -49,9 +49,9 @@ export class DgraphDataSource extends RESTDataSource {
           type
           value
         }
-        form: ~has_child {
-          category:  ~has_child {
-            product: ~has_child {          
+        parent: ~has_child {
+          parent:  ~has_child {
+            parent: ~has_child {          
               properties: has_property {
                 type
                 value
--- a/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
@@ -41,14 +41,24 @@ export class DgraphDataSource extends RESTDataSource {
         total: count(uid)
       }
       
-      query(func: uid(all), ${orderString}, offset: ${offset}, first: ${first}) @recurse(loop: false)  {
+      query(func: uid(all), ${orderString}, offset: ${offset}, first: ${first})  {
         code
         description
         type
-        uid
-        value
-        children: has_child
-        properties: has_property
+        properties: has_property {
+          type
+          value
+        }
+        form: ~has_child {
+          category:  ~has_child {
+            product: ~has_child {          
+              properties: has_property {
+                type
+                value
+              }
+            }
+          }
+        }
       }
     }`;
   }
@@ -84,7 +94,7 @@ export class DgraphDataSource extends RESTDataSource {
     const { counters: countersData, query: entityData } = data ?? {};
     const [counterData] = countersData;
     const totalCount = counterData?.total;
-    
+
     // Overwrite interactions to prevent large query delays.
     const entities: IEntity[] =
       entityData?.map((entity: IEntity) => ({ ...entity, interactions: [] })) ?? [];
--- a/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
@@ -26,6 +26,7 @@ export class DgraphDataSource extends RESTDataSource {
         value
         children: has_child
         properties: has_property
+        parent: ~has_child 
       }
     }`;
   }
--- a/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
@@ -27,7 +27,6 @@ export class DgraphDataSource extends RESTDataSource {
         value
         children: has_child
         properties: has_property
-        parent: ~has_child 
       }
     }`;
   }
--- a/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
@@ -7,6 +7,7 @@ import {
   IEntity,
   IEntityCollection,
 } from '@unified-codes/data';
+import { EntitySearchInput, FilterMatch } from '../schema';
 
 export class DgraphDataSource extends RESTDataSource {
   private static headers: { [key: string]: string } = {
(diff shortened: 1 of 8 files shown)


unified-codes/apps/data-service/src/app/types/DgraphDataSource.ts
```diff
@@ -64,14 +65,24 @@ export class DgraphDataSource extends RESTDataSource {
...
+        parent: ~has_child {
+          parent:  ~has_child {
+            parent: ~has_child {          
+              properties: has_property {
+                type
+                value
+              }
+            }
+          }
+        }
```
wlthomson, line 85: I think there's a more elegant way to do this in dgraph by using `@cascade`  (allows you to "bubble up" filters at lower levels to higher levels, i.e. "get properties for node which has child with code `x`"). Think best is if I do a demo in an issue rather than request changes to the PR 😨 .

unified-codes/libs/data/src/lib/types/Entity.ts
```diff
@@ -83,6 +104,34 @@ export class Entity implements IEntity {
...
+
+      if (parentProperty && !childProperty) {
+        if (!this.properties) {
+          this._properties = [new Property(parentProperty)];
+        } else {
+          this._properties.push(new Property(parentProperty));
+        }
+      }
+    });
+  }
```
wlthomson, line 134: A little confused by this also... I think the niceness of the recursive structure is that we can traverse the tree? Since you've exposed `parent` as a getter, aren't these methods redundant (e.g. we don't have `getChildProperties`, as can just do `child.properties`)?

wlthomson (approved):
Approving with reservations to be addressed in follow up issue 😨 .

I think this approach is baking in logic a bit unnecessarily. By exposing the `parent` as well as parent properties/mutations directly on the child, I feel like there's a bit of duplication going on? I think it would be good if we had a design principle which encouraged traversing the tree rather than us exposing it via application-specific methods (does that make sense)? E.g. if a user wants to get a parents properties, that's quite easy with `node.parent.properties`, as opposed to `node.getParentProperties`, which feels a bit overkill?

On a similar theme, I also think we are hardcoding the queries a bit too much, as opposed to making use of dgraph query functionality designed for this sort of stuff. With cascading filters, we can make the backend queries a lot simpler (although graphql doesn't support cascade, so we will still possibly need to do some clever resolver logic here).

@katherine-sussol @mark-prins Does this make sense? Agreed or no?

unified-codes/libs/data/src/lib/types/Entity.ts
wlthomson: Also double confused about `updatePropertiesFromParent`... what's the need to duplicate the properties on the child? As long as our schema is documented clearly, shouldn't the user know where the properties they want are located in the hierarchy?

unified-codes/libs/data/src/lib/types/Entity.ts
```diff
@@ -21,20 +28,23 @@ export interface IEntity {
   interactions?: IDrugInteraction[];
   children?: IEntity[];
   properties?: IProperty[];
+  parent?: IEntityParent[];
```
katherine-sussol, line 31: I kind of am also confused/uncertain about this new type - technically, in this case, the shape of this object would be the same as `IEntity`? Or is there something special about this one? 🤔

unified-codes/libs/data/src/lib/types/Entity.ts
```diff
@@ -64,10 +78,17 @@ export class Entity implements IEntity {
...
-    const [property] = this.properties?.filter((property: IProperty) => property.type === type) ?? [];
+    const [property] =
+      this.properties?.filter((property: IProperty) => property.type === type) ?? [];
     return property;
   }
 
+  getParentProperty(type: string): Property {
+    const properties = this.getParentProperties();
+    const [property] = properties?.filter((property: IProperty) => property.type === type) ?? [];
+    return property ? new Property(property) : undefined;
```
katherine-sussol, line 89: Something is showing up for me on this line: `TS2322: Type 'Property | undefined' is not assignable to type 'Property'.`

unified-codes/libs/data/src/lib/types/Entity.ts
```diff
@@ -83,6 +104,34 @@ export class Entity implements IEntity {
...
+  updatePropertiesFromParent() {
+    productPropertyTypes.forEach((type) => {
+      const parentProperty = this.getParentProperty(type);
+      const childProperty = this.getProperty(type);
+
+      if (parentProperty && !childProperty) {
+        if (!this.properties) {
+          this._properties = [new Property(parentProperty)];
+        } else {
+          this._properties.push(new Property(parentProperty));
```
katherine-sussol, line 130: And something also on this line: `TS2532: Object is possibly 'undefined'.`

katherine-sussol (approved):
Approving as I think the general idea of exposing the parent node is good 👍 

Few minor comments - will read and respond to the larger comment in another comment

mark-prins:
@wlthomson @katherine-sussol ...Ok, how's this approach?

I've added support for the client to request a `product` of an element, and then the client can handle the storing and usage of the `who_eml` property themselves.

Will work on drug interactions after this.

wlthomson (approved):
Hmm, sorry again, but I think this is binding UC to mSupply in a not so good way. Since an `Entity` can be higher up in the hierarchy than a product, this field doesn't really make sense as a general field* (it also kind of breaks from the nice pattern of having recursion so we don't have to explicitly encode the hierarchy). 

in general I think the best way to do this is to avoid adding new fields to the `Entity` class and find a way of doing it by passing a bit more responsibility over to the client query. 

Since this is `v1`, I think we can go with it and revisit in `v2`?

mark-prins:
> Since this is v1, I think we can go with it and revisit in v2?

I need this in mSupply now, so nope. Would like to have at least the shape of the query determined.

wlthomson:
> Approving as I think the general idea of exposing the parent node is good 👍

Looks like I'm outnumbered here, but I think having a two-way link is redundant and could cause data integrity issues! Since the parent-child relationship is already encoded by the parent, storing it again on the child seems unnecessary? I think we possibly are still thinking a bit too "RESTful" in terms of how we are exposing our graph via queries. Ideally I think we should be letting the client traverse the different levels, rather than setting up predefined queries as "endpoints" for the clients to use for specific things. IMO we shouldn't need to traverse "up" the hierarchy, as the client can specify this using the existing child relationships (in dgraph with reverse edges and `cascade`, and in graphql with either a bit of cleverness in our resolvers, or potentially with multiple queries... I know, yuck, but I still think the better option).

Does that make sense?

mark-prins:
do you mean that the client should query something like this? 
```
{
  query(func: eq(code, "528d0c00")) {	
  	code
    description
    properties: has_property {
    	type
      value
    }
    form: ~has_child {
     category: ~has_child {
        product: ~has_child  {
          description
          type
    			properties: has_property {
            type
            value
          }
        }
      }
    }
  }
}
```

katherine-sussol:
> > Approving as I think the general idea of exposing the parent node is good 👍
> 
> Does that make sense?

Hmm, I probably wasn't clear what I meant specifically by exposing the parent. I think we're on the same page as I otherwise agree with everything you've said. My statement was more around the fact that there was previously no ability to traverse up (i.e. node.parent wasn't defined prior to this PR). Not advocating for `Parent` as a specific endpoint or property on the child if that makes sense.

wlthomson:
> do you mean that the client should query something like this?

Not totally sure of how to do it in vanilla graphql, but in graphql+- would be something like what we got working earlier, e.g.:

```
{
  query (func: has(code)) @cascade { 
    properties {
      type
      value
    }
    has_child {
      has_child  {
        has_child @filter(alloftext(code, "46bf643e")) {
          code
          type
          description
          properties {
            type
            value
          }
        }
      }
    }
  }
}
```

There's probably a way to be clever here and aggregate the `properties`, but I think it's a lot easier to just assume anyone using the graphql directly is either us or other devs (or highly tech literate users). Make more sense?

mark-prins:
yep, all good thanks. The other fun directive I was playing with is `@normalize` to remove our nesty-ness.
As you say, good enough for v1

PR #308, opened 2020-12-03
#304 add interface to copy entity code to clipboard

### LINKED ISSUE #304: Can't copy the code of an item
In the web interface there's no way to copy a code - all this looking up of codes and no way other than transcribing to actually grab one (apart from the API of course ;-)  )

wlthomson:
Fixes #304.

Branches off #305.

## Description

Adds a button to `ExplorerTable` to copy entity codes to clipboard.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


katherine-sussol (approved):
Looks good, I'm no UX expert but seems pretty intuitive to me. Working well on my end - some minor comments:

- Should we include the `package.json` update? (`clipboard-copy`)
- Minor type script warning (I think because we added new entity type that's not available on the toggle bar) 🤷  
image

wlthomson:
> Looks good, I'm no UX expert but seems pretty intuitive to me. Working well on my end - some minor comments:
> 
> * Should we include a `package.json` update? (`clipboard-copy`)
> * Minor type script warning (I think because we added new entity type that's not available on the toggle bar - I don't even think this is related to this PR but the previous one) 🤷
>   image

Yeah, the whole confusion between types and categories in `v1` might be something we want to address... I'm not sure exactly how to do it without breaking mSupply backwards compatibility though!

PR #311, opened 2021-02-15
#310 issues with new data

### LINKED ISSUE #310: New data not compatible with v2 import
Some problems exist with the v2 import and display of data

- import does not handle any properties
- existing properties need to be retained
- name of items on the product display page is incorrect: shows only the strength value
- the icon is not showing in the child list
- synonyms are not used

mark-prins:
Fixes #310 

## Description
Gets it all working again 😄 
Well, almost.

- import handles RxNav - other properties to be added in separate issues
- name of items on the product display page has been updated
- the icon is not showing in the child list
- linking of items to children was not working

Overall - running `./setup.sh` now populates dgraph and the site displays much the same info as the current production site.

### Documentation:

- [ ] Documentation OK.

### Tests:

- [ ] Tests OK.


katherine-sussol (commented):
Nice! Some funky things going on with the API but I guess they aren't critical, the really important queries we care about for now would be the Web UI and mSupply ones which seem to be working fine 😆 . Could still create an issue to investigate maybe?

One note about the combinations mapping otherwise I think all good.

e.g. 'valid' queries are being rejected such as: 
```
query GetEntityByCode {
  entity (code: "C7750265") {
    code
    description
    type
    properties {
      type
      value
    }
    # product
    product {
      code
      type
      description
    }
    # form_category
    children {
      code
      description
      type
      # form
      children {
        code
        description
        type
        # unit_of_use
        children {
          code
          description
          type
          properties {
            type
            value
          }
        }
      }
    }
  }
}
```

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
```diff
@@ -198,6 +205,10 @@ export class DgraphDataSource extends RESTDataSource {
       return { ...entity, type, children };
     };
 
+    entity.properties = entity.properties?.map((property) => ({
```
katherine-sussol, line 208: Oh, sorry I lied, I think there is a small problem with the mSupply query - `getEntities` also needs this mapping I think (as well as `getEntity`)!

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
```diff
@@ -94,10 +96,21 @@ const DetailEntityListItemComponent: DetailEntityListItem = ({
...
+      let fullDescription = description;
+      let parentDescription = entityDescription;
+      switch(type) {
+        case 'form':
+          parentDescription = `${entityDescription} ${description}`;
+          break;
+        case 'unit_of_use':
+          fullDescription =  `${entityDescription} ${description}`;
+          break;
+      }
```
wlthomson, line 109: If I understand this right, this propagates higher level descriptions down to lower levels (and then concatenates them). I can understand the motivation, but the solution here breaks the purity of the recursion and makes it a bit confusing (at least to me!). Would an alternative (and possibly simpler) idea be to update the data (either at the `.csv`/spreadsheet stage or just moving this logic out of the UI and into the pre-processing), rather than prop drilling the parent's description? The benefit of manually setting the descriptions in the spreadsheet is that it avoid annoying edge-cases and such which could bloat this logic more and more (I'm guessing there will be situations where concatenating the descriptions might not give the desired effect)?

unified-codes/tools/scripts/dgraph/migration/v2/DataParser.ts
```diff
@@ -1,71 +1,57 @@
...
-    name?: string;
-    type?: string;
-    combines?: INode[];
-    properties?: INode[];
-    children?: INode[];
+  code: string;
+  name?: string;
+  type?: string;
+  combines?: INode[];
+  properties?: IProperty[];
```
wlthomson, line 39: Oops, maybe some confusion here. This is a "generic" type (to allow for the most flexible recursion). I think a more elegant/consistent solution could be to add `value?: string;` to `INode` and to revert properties back to `INode[]`.

Does that make sense?

unified-codes/tools/scripts/dgraph/migration/v2/DataParser.ts
```diff
@@ -75,413 +61,477 @@ export abstract class DataParser {
...
+      stream
+        .pipe(csv())
+        .on('data', (row) => {
+          const entity = Object.entries<string>(row).reduce(
+            (acc: IRow, [column, value]: [string, string]) => {
+              const key = column
+                .trim()
+                .toLowerCase()
+                .replace(/ *\([^)]*\) */g, '')
+                .replace(/ /g, '_');
```
wlthomson, line 181: Perhaps we could define these regex with a constant? Or just a comment would be fine, too!

wlthomson (changes requested):
Nice! A couple of implementation things, but mostly LGTM!

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
mark-prins: Thanks! d636da195

unified-codes/tools/scripts/dgraph/migration/v2/DataLoader.ts
```diff
@@ -1,118 +1,163 @@
...
-            for await (const sibling of entity.combines) {
-                const req = new dgraph.Request();
-                const query = `query {
+    // Link products to combinations.
+    for await (const entity of entities.filter((entity) => !!entity.combines)) {
+      for await (const sibling of entity.combines) {
+        const req = new dgraph.Request();
+        const query = `query {
                     Entity as var(func: eq(dgraph.type, ${entity.type})) @filter(eq(code, ${entity.code}))
                     Sibling as var(func: eq(dgraph.type, ${sibling.type})) @filter(eq(code, ${sibling.code}))
```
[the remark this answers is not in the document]
mark-prins: Thanks @katherine-sussol!
The rejected query has been fixed: 81613de06
Main problem is that the codes are now all lowercase, and `eq` is case sensitive. I've transformed in the query, and handled the no results scenario, so that we can return a null result.

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
```diff
@@ -175,6 +176,12 @@ export class DgraphDataSource extends RESTDataSource {
     }
   }
 
+  private static getPropertyType(property: IProperty): string {
+    const { type: types } = property;
+    const [type] = types ?? [];
+    return type;
+  }
```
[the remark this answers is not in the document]
mark-prins: `IProperty` is incorrect - with the change to use `dgraph.type` the `type` is now an array, which is why this change was needed.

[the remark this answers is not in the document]
mark-prins: Added d02119336 to rename `dgraph.type` to `types` and map that to `type` in `IProperty`
It's not great, but it is clearer.

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
mark-prins: 😄  yes, it's awful.
Are you suggesting we return to having a complete description on each entity rather than splitting out all the elements?

unified-codes/tools/scripts/dgraph/migration/v2/DataParser.ts
mark-prins: Makes sense. though it is in contrast to the rest of this codebase where we've used specific interfaces rather than using generics. 
Changed and ran a quick test, seems to be ok.  5d807f342

unified-codes/tools/scripts/dgraph/migration/v2/DataParser.ts
mark-prins: I like constants. Though I also had to comment the really dodgy one.
61e859237

unified-codes/tools/scripts/dgraph/migration/v2/DataLoader.ts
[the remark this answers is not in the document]
mark-prins: The combinations: another instance of an original bug. 
there was this: `this.graph[uc2].combines.push({ code: uc });`
and then 
```
for await (const sibling of entity.combines) {
...
Sibling as var(func: eq(dgraph.type, ${sibling.type})) @filter(eq(code, ${sibling.code}))```

given the way loader is working.. there is no `type` on `sibling` so new entities are created instead of looking up the correct ones.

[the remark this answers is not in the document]
mark-prins: Fixed in 1a018398e ( other branch )

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
mark-prins: Selectors!
I thought maybe reformatting as selectors would be nice; `getEntityDescription( code )` could return a fully formatted description

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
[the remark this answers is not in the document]
wlthomson: Now I get it! I think it's better the way you had it before. A comment would be plenty good enough, I think (rather than adding another `types` field, which is arguably a bit more confusing!). Sorry for the extra work 😢 !

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
wlthomson: Yeah, that way the `description` can be passed directly into the component. I think `selectEntityDescription` would be the preferred way of doing it, for consistency with other selectors (see `src/selectors/DetailSelectors`).

unified-codes/tools/scripts/dgraph/migration/v2/DataParser.ts
wlthomson: Yeah, I think the idea was to be as flexible as possible, since the `csv` format was so fluid. Also I think my idea was that it was good to distinguish the types representing the structure of the input file from the final processed types.

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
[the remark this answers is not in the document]
mark-prins: Now who's messing with whom??! this switch over to use `dgraph.type` and therefore an array of types was your move originally!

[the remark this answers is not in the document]
mark-prins: Revert and comment in 174db11fe

mark-prins:
> Nice! Nit: could probably put these right at the top? Also, for consistency with other constants:

174db11fe

mark-prins:
@wlthomson thanks again - should be ready now.

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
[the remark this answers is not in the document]
wlthomson: > 
> 
> Now who's messing with whom??! this switch over to use `dgraph.type` and therefore an array of types was your move originally!

https://youtu.be/UmzsWxPLIOo.

wlthomson (approved):

PR #327, opened 2021-03-10
#322 add product entity graphql mapping

### LINKED ISSUE #322: Product field resolver returns unmapped properties
Entities resolved by the `product` field resolver are returned with an incorrectly formatted `type` field. Fixing this bug is a prerequisite to fixing https://github.com/sussol/msupply/issues/8211.

## Expected Behaviour

Entity types should resolve to strings (e.g. `"code_rxnav"`).

## Current Behaviour

Product entity types are not mapped to strings (e.g. `["code_rxnav"`]).

## Possible Solution

Longer-term, I'm not sure of the value of a `property` field/field resolver (the client should be able to specify what they want to retrieve with a graph query, and shouldn't need to rely on customised resolver "endpoints"). Short term, the easiest solution is to fix the existing field resolver logic using existing helper methods (e.g. `DgraphDataSource.mapEntity`).

1. POST a GraphQL request to the `data-service` which requires resolving product property values (e.g. `{ product { properties { type value } } }`).
2. See error.

wlthomson:
Fixes #322.

## Description

Maps `product` entity property type values to the format specified by the graph schema.

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.

### DIFF
--- a/unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
@@ -223,7 +223,8 @@ export class DgraphDataSource extends RESTDataSource {
 
     const { query } = data ?? {};
     const [entity] = query ?? [];
-    return entity;
+
+    return DgraphDataSource.mapEntity(entity);
   }
 
   async getEntities(


katherine-sussol (approved):
LGTM (without really understanding the context of what product is used for)

PR #351, opened 2021-03-18
#350 add v2 entity constants

### LINKED ISSUE #350: Add v2 entity constants
Add `v2` entity constants to `v1` `data-service` mappers.

## Expected Behaviour

Hardcoded strings are represented using constants.

## Current Behaviour

E.g.

```
  private static getEntityTypeString(types: string[]) {
    return JSON.stringify(types.map(type => {
      switch (type) {
        case EEntityType.FORM_CATEGORY:
        case EEntityType.FORM:
          return 'Form';
        case EEntityType.UNIT_OF_USE:
        case EEntityType.STRENGTH:
          return 'Unit';
        case EEntityType.DRUG:
        default:
          return 'Product';
      }
    }));
  }
  ```

## Possible Solution

Import `@unified-codes/data/v1` entity constants.

### PR DESCRIPTION
**Pull request by wlthomson**
Fixes #350.

## Description

Replaces some hardcoded strings with clearer constants.

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.

### DIFF
--- a/unified-codes/libs/data/src/lib/v2/index.ts
+++ b/unified-codes/libs/data/src/lib/v2/index.ts
@@ -1,0 +1,2 @@
+export * from './entity';
+export * from './property';
--- a/unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
@@ -5,10 +5,17 @@ import {
   EEntityType,
   EEntityField,
   IEntity,
+  EEntitySortOrder,
   EntityCollection,
   IEntityCollection,
   IProperty,
 } from '@unified-codes/data/v1';
+
+import { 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: unified-codes/libs/data/src/lib/v2/index.ts [hunk 1]
```diff
@@ -0,0 +1,2 @@
+export * from './entity';
```
katherine-sussol, line 1: First letter 🔡  -> 🔠 ?

katherine-sussol (commented):
Liking the new constants 😎  - I'm having a bit of trouble getting this running, would it be better to just review #352 and merge that one as it looks like it has some fixes that get it working again? I kinda just realised a lot of my comments while I was adding them are resolved in that PR 😆

wlthomson:
As suggested, going to go ahead and batch merge this with #352 (as I think all the comments here are resolved in that PR!).

PR #362, opened 2021-03-24
#360 update data so all items have strength or unit of use

### LINKED ISSUE #360: Update master spreadsheet so all items have strength or unit of use
Required for https://github.com/sussol/msupply/issues/8272

## Expected Behaviour

All existing items on https://universalcodes.msupply.org.nz/ should be searchable in v2.0.0 of UC 

## Current Behaviour

Some items are missing strength or unit of use codes which means the item won't show up in mSupply

## Possible Solution

Update all items such that there is either a strength or unit of use present (for a lot of consumables, this will be '1')

This will be done by transferring UC2 to UC7 for these items, and then generating new UC2 codes for them.

1. Search in mSupply for any of the items in this spreadsheet (in the universal codes search box):
[Items without strength or unit 19 March.xlsx](https://github.com/openmsupply/unified-codes/files/6193180/Items.without.strength.or.unit.19.March.xlsx)
2. Observe no results returned

katherine-sussol:
Fixes #360 

## Description

- Updated products.csv with new data format (so all items have strength and/or unit of use)
- Updated data parser to correctly load these new nodes
- Updates to mSupply query so items show up in mSupply as expected
- Some minor offroads (will be noted in comments)

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.


unified-codes/apps/data-service/src/app/v1/schema.ts
```diff
@@ -19,6 +19,10 @@ export class EntityType implements IEntity {
   @Field((type) => [EntityType], { nullable: true })
   children: IEntity[];
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  @Field((type) => [EntityType], { nullable: true })
+  parents: IEntity[];
```
katherine-sussol, line 24: This was added with the intention to later remove `product` from the schema.

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
```diff
@@ -156,9 +156,11 @@ export class DgraphDataSource extends RESTDataSource {
             description: name@*
           }
           children {
+            code
```
katherine-sussol, line 159: Some codes were missing, guessing maybe it wasn't possible before but with all these new level mappings it's now possible? 🤷  was causing some items in mSupply to not show up

unified-codes/apps/web/.storybook/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../../tsconfig.base.json",
   "compilerOptions": {
+    "jsx": "react",
```
katherine-sussol, line 4: Looking back I think I was getting some warnings and made this change, can remove if it was just me

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
```diff
@@ -104,6 +104,7 @@ const DetailEntityListItemComponent: DetailEntityListItem = ({
           parentDescription = `${entityDescription} ${description}`;
           break;
         case 'unit_of_use':
+        case 'strength':
```
katherine-sussol, line 107: Previously the lowest level was only 'unit_of_use' but now we can go down to strength, was causing entity descriptions to appear truncated

unified-codes/apps/web/src/components/containers/detail/DetailProduct.tsx
```diff
@@ -64,8 +64,8 @@ export const DetailProductComponent: DetailProduct = ({ entity, fetchEntity }) =
         Product
       </Typography>
       <Typography className={classes.product} onClick={onProductClick}>
-        <Typography>{description}</Typography>
-        <Typography className={classes.code}>{code}</Typography>
+        {description}
```
katherine-sussol, line 67: Got JSX validation error about nested `<p>` tags. I think this is going to be removed anyway soon

unified-codes/apps/web/src/sagas/DetailSaga.ts
```diff
@@ -44,6 +44,18 @@ const getEntityQuery = (code: string) => `
       type
       description
     }
+    parents {
```
katherine-sussol, line 47: Intended as a replacement for product fetching. Happy to remove if objections 🤣

katherine-sussol:
Bit stale now so there are a lot of merge conflicts... taking a 👀

unified-codes/apps/web/src/sagas/DetailSaga.ts
mark-prins: can an entity have multiple parents? wondering why this is an array

wlthomson: As of recent changes, the `product` field  has been made redundant... so might be able to get rid of this entirely anyway? Although, to keep the GraphQL as flexible as possible, just having this field which exposes the reverse edge for `children` is probably a good idea (I don't think there are any entities with multiple parents, but since it's an array in the Dgraph data model, may as well keep it that way for simplicity and future-proofing)?

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
wlthomson: mSupply is now directly grabbing entities at the `Unit` and `DoseStrength` levels, so I think all this stuff is no longer needed!

unified-codes/apps/web/src/sagas/DetailSaga.ts
wlthomson: Also just a note that this isn't needed right now, as the web UI only traverses downwards. We definitely want to be able to see parents, and I think updating the `DetailPage` to pre-expand the list to show the entity and it's parents is the easy solution to this. This will require this field, so that's why I don't have an objection to keeping it in there even though we won't be using it right now.

unified-codes/apps/web/.storybook/tsconfig.json
katherine-sussol: Already done in a different PR so this is gonezers after merge

unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
katherine-sussol: Yup this stuff is all gone after merging in `master` 🥳

unified-codes/apps/web/src/components/containers/detail/DetailEntityListItem.tsx
katherine-sussol: Component no longer does specific type mapping so this change has also gone after merging in `master`

unified-codes/apps/web/src/components/containers/detail/DetailProduct.tsx
katherine-sussol: No more DetailProduct.tsx

katherine-sussol:
Merge done - ready again for review 👍

unified-codes/apps/web/src/sagas/DetailSaga.ts
katherine-sussol: Hm yeah, I don't think there is currently a use case for multiple `parents` (although there will potentially be multiple levels of `parent`s above any given node). The schema does allow it though because it's a direct reversal of `children` which is an array 🤔 . I guess we could add parents manually as a separate edge but thought using reverse edge would be simpler 😆

wlthomson (approved):
LGTM!

Nit: for another issue, but I think also need to add some conditional logic to the web UI:

image

Issue #396, opened 2021-05-17
Add visual indication of ** wildcard?


DhanyaHerath:
## Expected Behaviour

In a [support ticket](https://support.msupply.org.nz/cerb6/profiles/ticket/TH-5195-TJ) Lauren (CHAI, Laos) asks whether you can search for an item that contains a phrase. 
![image](https://user-images.githubusercontent.com/29744386/118465740-4a817280-b756-11eb-9bc4-13c362af2b22.png)

Ideally we could later add a toggle or some way of changing this in the UI, but until then, do you think it could be useful to have some helper text indicating how it works?

## Current Behaviour

No indication of how to search for "includes x" rather than "starts with x"

katherine-sussol:
Hmm thinking of something like:
![image](https://user-images.githubusercontent.com/65875762/119279831-4e475480-bc82-11eb-8784-f0d27fa347d1.png)

or 

![image](https://user-images.githubusercontent.com/65875762/119279894-99616780-bc82-11eb-84a9-cedf4774c179.png)

mark-prins:
I like that - but I worry that it is more useful for developers and 'power users'
Do normal people search using wildcard indicators?

If that was in place, there would have to be more information - telling people what the asterisk means.
I would go with a more clumsy and failsafe approach
Either
- open out the search, so that is always searches with wildcards ( probably will return too many results )
- have a checkbox or other UI element to enable the full search and default to 'starts with' ( the old site functionality )

katherine-sussol:
It sounds like some people do use `*`(I think it's really an unintended feature because the search string just gets chucked straight into a regex where it just happens to work, Rowena found it with her extensive testing 🤣 ). Most useful I think for combination drugs where users want to find all items that contain a particular drug (e.g. *lami will give Abacavir/Lamivudine)

I do like the UI element idea though (just wary of changing _tooo_ much haha), maybe radio buttons under the text input with `startsWith` and `contains` options?

DhanyaHerath:
I promised Craig this set of work would take under a week :P How long would radio buttons take to implement?
Or could the "eg" text be changed to be `e.g. amoxicillin, amox, or *mox to search for items containing "mox"` or is that too clunky...? (I like the first option from [this comment](https://github.com/openmsupply/unified-codes/issues/396#issuecomment-846642008) by the way, I think it looks nicer!)

katherine-sussol:
[...]

PR #399, opened 2021-05-21
#387 expand product tree by default

### LINKED ISSUE #387: It should be easier to see all available product strengths in the Web UI
User feedback indicates that having to click a 'down arrow' toggle multiple times to view all available product strengths is a bit cumbersome.

This arrow:
image

## Expected Behaviour

An easy way to see all product strengths

## Current Behaviour

An easy but kind of annoying requirement to click on lots of arrows

## Possible Solution

- A good intermediary would be to either default to the whole product tree expanded, or an option in the UI to 'expand all'.
- Longer-term solution would probably require bit of thought and rework on UI design

## Jobcode 
MFAT TC2

katherine-sussol:
Fixes #387

## Description

- Expand entity tree by default to make it easier for users to see all available codes for a particular product.
- **Bold** the description and code for child-most items to make it clearer which items are able to be dispensed. 
  - Not convinced it's super clear, but we couldn't really think of a better idea (icons would probably be even more confusing, and removing the copy icon for higher-level codes seems a bit mean since some people might want the higher codes for other use cases that aren't mSupply?)
- Ran prettier over the changed file (soz I probably just made it harder to review 😞 )  

image

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.

### DIFF
--- a/unified-codes/apps/web/src/components/containers/detail/DetailEntityList.tsx
+++ b/unified-codes/apps/web/src/components/containers/detail/DetailEntityList.tsx
@@ -58,7 +58,7 @@ export type DetailEntityList = React.FunctionComponent<DetailEntityListProps>;
 export const DetailEntityListComponent: DetailEntityList = ({ description, parent, entities }) => {
   const classes = useStyles();
 
-  const { isOpen, onToggle } = useToggle(false);
+  const { isOpen, onToggle } = useToggle(true);
 
   const entityCount = entities.length;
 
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
The bold certainly helps!
Sorry to see that the `<FormIcon />` was removed at some point, I liked having those icons. 

Agree that the layout needs some help - the expanded view is better though. I found the many-clicks approach annoying to use. We can look at the layout another time :)

DhanyaHerath:
Exciting!!!! I got overly excited and went to check the website - guessing it's not up yet though?

katherine-sussol:
Hehe 😹  not yet, was thinking to upload once #396 is done?

PR #400, opened 2021-05-24
#396 add visual indication of wildcard

### LINKED ISSUE #396: Add visual indication of ** wildcard?

katherine-sussol:
Fixes #396

## Description

Add some placeholder and helper text to the explorer UI
Went with @mark-prins  suggestion here: https://github.com/openmsupply/unified-codes/issues/396#issuecomment-846656707
image

I got a bit confused with where to put the props and helper text (also the placeholder text for a code search seems a bit redundant but the filter type is there so...just in case someone tries to use it I guess?)

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.

### DIFF
--- a/unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
+++ b/unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
@@ -66,7 +66,7 @@ export const ExplorerPageComponent: ExplorerPage = ({
       classes={classes}
       table={<ExplorerTable />}
       toggleBar={<ExplorerToggleBar />}
-      searchBar={<ExplorerSearchBar />}
+      searchBar={<ExplorerSearchBar helperText={'Use * to indicate any characters'} />}
       progressBar={<ExplorerProgressBar type='material-ui' />}
     />
   );
--- a/unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
+++ b/unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
@@ -66,7 +66,7 @@ export const ExplorerPageComponent: ExplorerPage = ({
       classes={classes}
       table={<ExplorerTable />}
       toggleBar={<ExplorerToggleBar />}
-      searchBar={<ExplorerSearchBar helperText="Use * to indicate any characters" />}
+      searchBar={<ExplorerSearchBar helperText="Use * in place of one or more letters at the start" />}
       progressBar={<ExplorerProgressBar type="material-ui" />}
     />
   );
--- a/unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
+++ b/unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
@@ -66,7 +66,7 @@ export const ExplorerPageComponent: ExplorerPage = ({
       classes={classes}
       table={<ExplorerTable />}
       toggleBar={<ExplorerToggleBar />}
-      searchBar={<ExplorerSearchBar helperText={'Use * to indicate any characters'} />}
+      searchBar={<ExplorerSearchBar helperText="Use * to indicate any characters"/>}
       progressBar={<ExplorerProgressBar type='material-ui' />}
     />
   );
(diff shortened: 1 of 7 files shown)


unified-codes/apps/web/src/components/containers/explorer/ExplorerSearchBar.tsx
```diff
@@ -2,17 +2,45 @@ import * as React from 'react';
...
 import { ITheme } from '../../../styles';
 
-const styles = (_: ITheme) => ({
-  input: { paddingLeft: 15 },
-  button: { marginTop: 15 },
-});
+const useStyles = makeStyles((theme: ITheme) =>
+  createStyles({
+    root: { color: theme.palette.text.primary },
+    input: { paddingLeft: 15 },
```
mark-prins, line 15:
```suggestion
    input: { paddingLeft: 15, color: '#a2a2a2' },
```

I find the white text strange

Screen Shot 2021-05-25 at 9 51 50 AM

and would suggest adding the colour here, and then updating `ClearInput` to use the style. That results in:

Screen Shot 2021-05-25 at 10 18 13 AM

and

Screen Shot 2021-05-25 at 10 18 20 AM

unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
```diff
@@ -66,7 +66,7 @@ export const ExplorerPageComponent: ExplorerPage = ({
       classes={classes}
       table={<ExplorerTable />}
       toggleBar={<ExplorerToggleBar />}
-      searchBar={<ExplorerSearchBar />}
+      searchBar={<ExplorerSearchBar helperText={'Use * to indicate any characters'} />}
```
mark-prins, line 69: As this is text, there's no need for `{'`
( I prefer the standard of double-quoting component parameters, though I see that the line below isn't, so you may prefer to go with consistency )

```suggestion
      searchBar={<ExplorerSearchBar helperText="Use * to indicate any characters"/>}
```

unified-codes/apps/web/src/selectors/ExplorerSelectors.ts
```diff
@@ -45,6 +45,17 @@ const selectLabel = createSelector(selectFilterBy, (filterBy: EEntityField): str
   }
 });
 
+const selectPlaceholder = createSelector(selectFilterBy, (filterBy: EEntityField): string => {
+  switch (filterBy) {
+    case EEntityField.CODE:
+      return 'e.g. 4231441c';
+    case EEntityField.DESCRIPTION:
+      return 'e.g. amoxicillin, amox, *mox';
```
mark-prins, line 53: what would be good is to explain that we default to 'starts with' - but I think there is enough instructional text already! I'd let people figure that out for themselves..

unified-codes/apps/web/src/components/containers/explorer/ExplorerPage.tsx
katherine-sussol: Oh good point :) thanks

unified-codes/apps/web/src/components/containers/explorer/ExplorerSearchBar.tsx
katherine-sussol: Thanks, the fading of the label was a bit odd, this is much nicer :) ee895df5a

unified-codes/apps/web/src/selectors/ExplorerSelectors.ts
katherine-sussol: Hmm. Yeah, I'm conflicted now. It _is_ a bit weird that we default to starts with although it also matches mSupply's weird behaviour...

katherine-sussol:
> Still not sure of the need to limit the classes here.
> I would use
> 
> ```
>       classes={classes}
>       InputLabelProps={{classes}}
> ```
> 
> for these. Either way works though

Thanks, went ahead and changed, easier on the eyes :).

I also edited the helper text a bit after some further thought (although ideally we'd also fix https://github.com/openmsupply/unified-codes/issues/338 and improve the search to use a toggle/checkbox/radio button/something else).

Issue #407, opened 2021-06-20
Generate UCS codes for vaccines


katherine-sussol:
## Expected Behaviour

Unique codes should exist for each of the vaccines in the spreadsheet here: https://docs.google.com/spreadsheets/d/1lk13CuxnhZoY0gUo2DdbZ6tggTyzhMgq/edit#gid=712916393

## Current Behaviour

No vaccines on the universal codes V2 server (with the exception of COVID ones)

## Possible Solution

- Cmd + Option + u in empty UC cells, copy paste as applicable
- Export to .csv
- Push to GitHub

Atterwards either need to update import script to import from vaccines tab or merge vaccines tab into general items tab.

Issue #357, opened 2021-03-22
Properties are assigned codes on parsing CSV


wlthomson:
When parsing `products.csv`, properties are assigned randomly generated codes.

## Expected Behaviour

Properties are just `type: value` key-pairs attached to parent entities.

## Current Behaviour

Codes are generated, but don't seem to be used (type definitions and schemas don't include).

## Possible Solution

Remove code generation for properties?

@mark-prins Don't want to move on this without checking with you first if there was a reason for this that I'm not aware of?

mark-prins:
as I remember it.. they were being generated because someone defined a schema which included mandatory code.
It was easier to generate codes where they were missing than make the code optional

PR #408, opened 2021-06-22
#358 Update data parser

### LINKED ISSUE #358: Modularise repeated DataParser logic
`DataParser` includes repeated logic, e.g.:

```
// If row includes pack size code...
if (uc9) {
    const code = uc9;
    const name = pack_size;
    const type = EEntityType.PackSize;

    // and node does not exist...
    if (!(uc9 in this.graph)) {
        // create pack size node.
        const node = {
            code,
            name,
            type,
            children: [],
            properties: [],
        };

        this.graph[uc9] = node;

        console.log(`INFO: Created node of type ${type}: ${JSON.stringify(node)}`);
    }
}
```

## Expected Behaviour

Add function for encapsulating repeated logic.

## Current Behaviour

`DataParser` is very big, ~`650` lines.

## Possible Solution

See above.

katherine-sussol:
Fixes #358 and #357

## Description

... Sorry I had to.
Was investigating/looking at #407 and it's just too tedious to troubleshoot missing codes/levels every time we need new ones 😿 

- Modularised node creation logic
- Modularised child linking logic
- Removed code generation from property nodes

### Documentation:

- [x] Documentation OK.

### Tests:

- [x] Tests OK.

### DIFF
--- a/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
+++ b/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
@@ -185,7 +185,15 @@ export class DataParser {
         ];
 
         productDefinition.forEach(item => {
-          if (item.code && !(item.code in this.graph)) {
+          if (!item.code) return;
+          
+          if (item.code in this.graph) {
+            // check for conflicts.
+            if (item.type && this.graph[item.code].type != item.type) {
+              duplicates.push(item.code);
+            }
+          }
+          else {
             const node = { 
               code: item.code,
               name: item.name,
@@ -197,12 +205,6 @@ export class DataParser {
             this.graph[item.code] = node;
 
             console.log(`INFO: Created ${item.type} node: ${JSON.stringify(node)}`);
-          }
-          else if (item.code) {
-            // check for conflicts.
-            if (item.type && this.graph[item.code].type != item.type) {
-              duplicates.push(item.code);
-            }
           }
         });
 
--- a/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
+++ b/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
@@ -170,45 +170,19 @@ export class DataParser {
           // { name: row.brand, code: row.uc12, type: EEntityType.Brand },
         ];
 
-        const {
-          uc1,
-          combination,
-          rxnav,
-          who_eml_product,
-          who_eml_item,
-          nzulm,
-          nzulm_item,
-          unspsc,
-        } = row;
+        const { uc1, combination } = row;
 
-        const productProperties: IPropertyNode[] = [];
-        const itemProperties: IPropertyNode[] = [];
+        const productProperties: IPropertyNode[] = [
+          { type: EPropertyType.RxNav, value: row.rxnav },
+          { type: EPropertyType.WHOEML, value: row.who_eml_product },
+          { type: EPropertyType.NZULM, value: row.nzulm },
+          { type: EPropertyType.UNSPSC, value: row.unspsc }
+        ];
 
-        productProperties.push({
-          type: EPropertyType.RxNav,
-          value: rxnav,
-        });
-        productProperties.push({
-          type: EPropertyType.WHOEML,
-          value: who_eml_product,
-        });
-
-        itemProperties.push({
-          type: EPropertyType.WHOEML,
-          value: who_eml_item,
-        });
-        productProperties.push({
-          type: EPropertyType.NZULM,
-          value: nzulm,
-        });
-        itemProperties.push({
-          type: EPropertyType.NZULM,
-          value: nzulm_item,
-        });
-        productProperties.push({
-          type: EPropertyType.UNSPSC,
-          value: unspsc,
-        });
+        const itemProperties: IPropertyNode[] = [
+          { type: EPropertyType.WHOEML, value: row.who_eml_item },
+          { type: EPropertyType.NZULM, value: row.who_eml_item },
+        ];
 
         productDefinition.forEach(item => {
           if (item.code && !(item.code in this.graph)) {
--- a/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
+++ b/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
@@ -248,7 +248,6 @@ export class DataParser {
         // This avoids having to manually define each combination of 'levels' that are able to be connected 
         let parentIndex = 0;
         let childIndex = 1;
-        console.log(productDefinition)
         while (childIndex < productDefinition.length) {
           const parent = productDefinition[parentIndex];
           const child = productDefinition[childIndex];
--- a/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
+++ b/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
@@ -232,7 +232,7 @@ export class DataParser {
           }
         });
 
-        const productCode = productDefinition[0].code;
+        const productCode = productDefinition.find(item => item.type === EEntityType.Product)?.code;
         // If category code exists, and product code exists
         if (uc1 && productCode) {
 
@@ -297,8 +297,8 @@ export class DataParser {
           });
         }
 
-        const strengthCode = productDefinition[4].code; // UC6
-        const unitCode = productDefinition[5].code;     // UC7
+        const strengthCode = productDefinition.find(item => item.type === EEntityType.DoseStrength)?.code;  // UC6
+        const unitCode = productDefinition.find(item => item.type === EEntityType.Unit)?.code;      // UC7
 
         // Process external properties at item (UC6) level
         if (!unitCode && strengthCode) {
--- a/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
+++ b/unified-codes/apps/data-loader/src/app/v2/DataParser.ts
@@ -99,12 +99,6 @@ export class DataParser {
     this.cycles = [];
   }
 
-  private generateCode(numCharacters = 8) {
-    return Math.round(Math.random() * Math.pow(16, numCharacters))
-      .toString(16)
-      .padStart(numCharacters, '0');
-  }
-
   public async parseData(): Promise<ICSVData> {
     if (this.isParsed) return this.data;
 
@@ -162,32 +156,23 @@ export class DataParser {
 
       // Parse entity graph.
       this.data.forEach((row) => {
+        const productDefinition = [
+          { name: row.product, code: row.uc2, type: EEntityType.Product },
+          { name: row.route, code: row.uc3, type: EEntityType.Route },
+          { name: row.dose_form, code: row.uc4, type: EEntityType.Form },
+          { name: row.dose_qualification, code: row.uc5, type: EEntityType.FormQualifier },
+          { name: row.strength, code: row.uc6, type: EEntityType.DoseStrength },
+          { name: row.unit_of_presentation, code: row.uc7, type: EEntityType.Unit },
+          { name: row.immediate_packaging, code: row.uc8, type: EEntityType.PackImmediate },
+          { name: row.pack_size, code: row.uc9, type: EEntityType.PackSize },
+          // { name: row.outer_packaging, code: row.uc10, type: EEntityType.PackOuter },
+          // { name: row.manufacturer, code: row.uc11, type: EEntityType.Manufacturer },
+          // { name: row.brand, code: row.uc12, type: EEntityType.Brand },
+        ];
+
         const {
-          product,
-          //product_synonym,
-          combination,
-          route,
-          dose_form,
-          dose_qualification,
-          strength,
-          unit_of_presentation,
-          immediate_packaging,
-          pack_size,
-          //outer_packaging,
-          //manufacturer,
-          //brand,
           uc1,
-          uc2,
-          uc3,
-          uc4,
-          uc5,
-          uc6,
-          uc7,
-          uc8,
-          uc9,
-          //uc10,
-          //uc11,
-          //uc12,
+          combination,
           rxnav,
           who_eml_product,
           who_eml_item,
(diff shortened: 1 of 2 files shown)


unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -162,32 +156,23 @@ export class DataParser {
...
+          { name: row.dose_qualification, code: row.uc5, type: EEntityType.FormQualifier },
+          { name: row.strength, code: row.uc6, type: EEntityType.DoseStrength },
+          { name: row.unit_of_presentation, code: row.uc7, type: EEntityType.Unit },
+          { name: row.immediate_packaging, code: row.uc8, type: EEntityType.PackImmediate },
+          { name: row.pack_size, code: row.uc9, type: EEntityType.PackSize },
+          // { name: row.outer_packaging, code: row.uc10, type: EEntityType.PackOuter },
+          // { name: row.manufacturer, code: row.uc11, type: EEntityType.Manufacturer },
+          // { name: row.brand, code: row.uc12, type: EEntityType.Brand },
+        ];
+
```
mark-prins, line 172: From lines 159:211 ( I can't select all in the PR.. you could make that all a bit neater now?
```suggestion
        const { uc1, combination } = row;
        const productDefinition = [
          { name: row.product, code: row.uc2, type: EEntityType.Product },
          { name: row.route, code: row.uc3, type: EEntityType.Route },
          { name: row.dose_form, code: row.uc4, type: EEntityType.Form },
          { name: row.dose_qualification, code: row.uc5, type: EEntityType.FormQualifier },
          { name: row.strength, code: row.uc6, type: EEntityType.DoseStrength },
          { name: row.unit_of_presentation, code: row.uc7, type: EEntityType.Unit },
          { name: row.immediate_packaging, code: row.uc8, type: EEntityType.PackImmediate },
          { name: row.pack_size, code: row.uc9, type: EEntityType.PackSize },
          // { name: row.outer_packaging, code: row.uc10, type: EEntityType.PackOuter },
          // { name: row.manufacturer, code: row.uc11, type: EEntityType.Manufacturer },
          // { name: row.brand, code: row.uc12, type: EEntityType.Brand },
        ];
        const productProperties: IPropertyNode[] = [
          { type: EPropertyType.RxNav, value: row.rxnav },
          { type: EPropertyType.WHOEML, value: row.who_eml_product },
          { type: EPropertyType.NZULM, value: row.nzulm },
          { type: EPropertyType.UNSPSC, value: row.unspsc },
        ];
        const itemProperties: IPropertyNode[] = [
          { type: EPropertyType.WHOEML, value: row.who_eml_item },
          { type: EPropertyType.NZULM, value: row.nzulm_item },
        ];
```

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -200,403 +185,83 @@ export class DataParser {
...
-            };
-
-            this.graph[uc8] = node;
-
-            console.log(`INFO: Created node of type ${type}: ${JSON.stringify(node)}`);
-          }
-        }
-
-        // If row includes unit of presentation code...
-        if (uc7) {
```
mark-prins, line 281: uc7, 8 & 9 now have the conflict / duplicate check. that's ok?

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -200,403 +185,83 @@ export class DataParser {
...
+          if (child.name && child.code) {
+            if (!this.graph[parent.code].children.map((child) => child.code).includes(child.code)) {
+              this.graph[parent.code].children.push(child);
               console.log(
-                `INFO: Linked dose qualifier with code ${uc5} to dose strength with code ${uc6}`
-              );
-            }
-          }
-          // if strength doesn't exist but unit of presentation does
-          else if (uc7) {
```
mark-prins, line 505: I'm not seeing where this has gone in the new version

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -200,403 +185,83 @@ export class DataParser {
...
-                `INFO: Linked dose qualifier with code ${uc5} to unit of presentation with code ${uc7}`
+                `INFO: Linked ${parent.type} with code ${parent.code} to ${child.type} with code ${child.code}`
               );
             }
+            parentIndex = childIndex;
           }
-        }
-
-        // If dose form code exists...
-        if (uc4) {
```
mark-prins, line 517: I can't see this in the new version

mark-prins (changes requested):
I think I understand it all.. and overall it is awesome!
Long overdue refactoring - and I've suggested a small additional compression.

There's some conditions which I think have been missed out though; the 'if this and this then do some stuff' sections. One of them has been migrated only ( `strengthCode && !unitCode` ) along with the `uc1 && productCode`. Don't the rest need to come along?

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -200,403 +185,83 @@ export class DataParser {
         const itemProperties: IPropertyNode[] = [];
 
         productProperties.push({
-          code: this.generateCode(),
```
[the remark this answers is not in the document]
mark-prins: seems ok without them 🤷‍♂️

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -200,403 +185,83 @@ export class DataParser {
...
-          if (uc6) {
-            // link dose qualification to strength.
-            if (!this.graph[uc5].children.map((child) => child.code).includes(uc6)) {
-              this.graph[uc5].children.push({ code: uc6 });
+        // Iterate through and assign children to parents 
+        // This starts at the top of the graph and for each node, finds the closest lower level node that has a code and name
+        // This avoids having to manually define each combination of 'levels' that are able to be connected 
+        let parentIndex = 0;
+        let childIndex = 1;
+        console.log(productDefinition)
```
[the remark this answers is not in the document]
katherine-sussol: Whoops 2ce58fa

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -200,403 +185,83 @@ export class DataParser {
...
-          // and pack immediate code exists...
-          if (uc8) {
-            // link unit to pack immediate.
-            if (!this.graph[uc7].children.map((child) => child.code).includes(uc8)) {
-              this.graph[uc7].children.push({ code: uc8 });
-              console.log(`INFO: Linked unit with code ${uc7} to pack immediate with code ${uc8}`);
-            }
-          }
-        }
+        const productCode = productDefinition[0].code;
```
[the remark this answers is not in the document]
katherine-sussol: Oh yes, that is much clearer/safer, thanks! 160708c52

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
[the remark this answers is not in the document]
katherine-sussol: It does now, although I'm not sure exactly if/when that was changed. At least at the moment the GraphQL schema only has code defined as a required field for `Entity` types, but not `Property` types (which just has the code stored in the `value` field).

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -200,403 +185,83 @@ export class DataParser {
...
-
-          // and node does not exist...
-          if (!(uc2 in this.graph)) {
-            // create product node.
-            const node = {
-              code,
-              name,
-              type,
+        productDefinition.forEach(item => {
+          if (item.code && !(item.code in this.graph)) {
```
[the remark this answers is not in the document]
katherine-sussol: Thanks, good call 😸  b4377e716

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
katherine-sussol: 🧹  a6724bba6

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
katherine-sussol: Hmm, I _think_ it should be ok - we don't really use/expose uc8 or uc9 at the moment - but my thinking was that they should still be unique regardless of level?

There are a few duplicates found with this update (I think maybe they should be updated in a separate issue)
- bb88f600
- 105be100
- 574f2000

I'm also a bit wary that this only checks the type...
If, say unit of use was the lowest code for two different products, and they happened to be assigned the same UC, it would probably still pass this check?

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
katherine-sussol: The idea was that this would / should handle each case of checking a node for its child:
https://github.com/openmsupply/unified-codes/blob/b4377e7168ebf83baaf8fcbd33360082a5c84cf5/unified-codes/apps/data-loader/src/app/v2/DataParser.ts#L225-L240

E.g if UC2-7 were all defined except for UC6 (strength) this block should do the following linkages  
- UC2 -> UC3 (`productDefinition[0]` and `productDefinition[1]`)
- UC3 -> UC4 (`productDefinition[1]` and `productDefinition[2]`)
- UC4 -> UC5 (`productDefinition[2]` and `productDefinition[3]`)
- `// UC6 skipped because there's no name and code defined for productDefinition[4]`
- UC5 -> UC7 (`productDefinition[3]` and `productDefinition[5]`)

... Hmm I'm not sure I'm explaining very well (I guess a unit test would also be good for a function like this! 🙀

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
mark-prins: yes - agree. I think it's fine to have the check, and we should probably be checking regardless of type. but fine for now.

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
katherine-sussol: _Shouuuuld_ be covered by the above - (happy for feedback on the crude logic though!)

katherine-sussol:
Have just re-requested a review as I _think_ all the main points have been addressed (I also don't think anyone's waiting on these vaccines right now but I guess as soon as someone does ask about it, it will be good to be ready with a loader that doesn't skip arbitrary nodes 😟 )

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -162,441 +156,87 @@ export class DataParser {
...
+        const productProperties: IPropertyNode[] = [
+          { type: EPropertyType.RxNav, value: row.rxnav },
+          { type: EPropertyType.WHOEML, value: row.who_eml_product },
+          { type: EPropertyType.NZULM, value: row.nzulm },
+          { type: EPropertyType.UNSPSC, value: row.unspsc }
+        ];
+
+        const itemProperties: IPropertyNode[] = [
+          { type: EPropertyType.WHOEML, value: row.who_eml_item },
+          { type: EPropertyType.NZULM, value: row.who_eml_item },
```
mark-prins, line 184: good ol' copy paste error :)
```suggestion
          { type: EPropertyType.NZULM, value: row.nzulm_item },
```

katherine-sussol: Ooops, good catch haha 🙀

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
```diff
@@ -162,441 +156,87 @@ export class DataParser {
...
+        // Iterate through and assign children to parents 
+        // This starts at the top of the graph and for each node, finds the closest lower level node that has a code and name
+        // This avoids having to manually define each combination of 'levels' that are able to be connected 
+        let parentIndex = 0;
+        let childIndex = 1;
+        while (childIndex < productDefinition.length) {
+          const parent = productDefinition[parentIndex];
+          const child = productDefinition[childIndex];
+          if (child.name && child.code) {
+            if (!this.graph[parent.code].children.map((child) => child.code).includes(child.code)) {
```
mark-prins, line 231: I would expect the redefinition of `child` to cause a problem here.
There's a `const child` in line 229 and now `children.map((child) =>...`
One of them will need to be renamed :)

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
mark-prins: We also have these cases in the current version which don't map to anything in the new that I can see:
- `uc2 && !uc3 && uc6` 
- `uc2 && !uc3 && !uc6 && uc7`
- `uc5 && !uc6 && uc7`

Since they've deliberately been put in place, I think we should reimplement, sorry!

mark-prins (changes requested):
A couple of minor things to update. 
The change is a huge improvement tho - its worth pressing on!

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
katherine-sussol: Thanks - I didn't notice that (interestingly they didn't seem to conflict at runtime?), but I agree it's super confusing even if it happened to work 🤣  21effa8eb

unified-codes/apps/data-loader/src/app/v2/DataParser.ts
katherine-sussol: Hmm, going to need a re-think on the algorithm, perhaps. What I've done I guess is a bit blind to the structure validation rules and relies on accurate spreadsheet entry, whereas the previous implementation has built-in assumptions about data structure rules.

With the existing logic, all of the above scenarios will be covered, but only if there aren't any codes in between.

E.g. For the first example
uc2 will still be mapped to uc6 if there is no 3, 4 or 5 defined.
If there is a uc4 (someone put it in the spreadsheet by accident), then the structure created will be uc2 -> uc4 -> uc6 which is technically invalid according to the SoP: 
- Items cannot have a dose form without a route:  https://docs.google.com/document/d/1xKmWT0Qiyam7tHBUTjZUteyF8VULOYFG/edit)  😢 

Maaaaybe, I should revert the tree builder and leave the old logic, it just hurts so much to read and it also skips nodes without warning 😆. Perhaps a better solution would be to log the skipped codes somewhere?

wlthomson (approved):
LGTM!

wlthomson:
@mark-prins 

> We also have these cases in the current version which don't map to anything in the new that I can see:

    uc2 && !uc3 && uc6
    uc2 && !uc3 && !uc6 && uc7
    uc5 && !uc6 && uc7

Apologies, been a while since I've been involved in UC. Are these "conditional" links, e.g. UC2 links to UC6 if no UC3 exists?

mark-prins:
> Are these "conditional" links, e.g. `uc2` links to `uc6` only if no `uc3` exists etc.? Wondering if we should document the graph topology logic somewhere?

Yeah they are. I don't know why they're in the code, just assumed that because they were added deliberately, there must be a reason.  Sounds like they _might_ be ok / catered for? Though what is mentioned is slightly different to what was in the code, in practice it might have the same result due to the data in the sheet.

mark-prins (approved):

PR #419, opened 2023-11-30
Kdd v3

lache-melvin:
## Description

Adds initial KDD document


jmbrunskill (approved):
Looks good. I think every thing we need to capture has been captured.

PR #426, opened 2023-12-06
424 update deps

### LINKED ISSUE #424: Clean up data-service warnings
Some love is required...

Image

Also, need to update dependencies... to run the data loader or the data service with a node version later than v16, you have to run with `NODE_OPTIONS=--openssl-legacy-provider`. We're up to node 20 now... be good to get everything up to date!

lache-melvin:
Fixes #424

## Description

Updates some dependencies, gets eslint working

- Apollo v3 is going to end of life next year, but v4 requires `graphql@16`, which we can't upgrade to until `type-graphql` releases support for it in their 2.0... it has been almost ready for release since the start of the year, so we might need to keep an eye on this? 
  - Also can't upgrade `fastify` until we have the new apollo fastify server
- Haven't updated any dependencies relating to `nx` or the existing frontend code, as this will all be removed
  - There are build warnings in the console from `nx`, as well as some typescript errors from the UI code... please just smile past them for a bit longer :)
  - We'll also need to keep either running with Node <16, or the `NODE_OPTIONS=--openssl-legacy-provider` until nx is removed, the `@nrwl` packages depend on legacy code
- `class-validator` could only be updated to 0.13 (not 0.14) due to a [change in how it handles unknown fields](https://github.com/typestack/class-validator/issues/1873) ... tried to troubleshoot but was taking too much time to figure out what the unknown fields were and update accordingly, should be handled as a separate issue I think


jmbrunskill (approved):
Works for me :)

PR #428, opened 2023-12-06
Fix error when querying non-existent entity

lache-melvin:
## Description

If query for entity has no result for whatever reason, we weren't checking for its existence before pulling children off it...

Screenshot 2023-12-07 at 11 46 08 AM

Adds a check and early return if `!entity` :)

### DIFF
--- a/unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
+++ b/unified-codes/apps/data-service/src/app/v1/types/DgraphDataSource.ts
@@ -78,6 +78,7 @@ export class DgraphDataSource extends RESTDataSource {
     return `{
       query(func: eq(code, ${code?.toLowerCase()}), first:1) @recurse(loop:false) {
         code
+        uid
         type: dgraph.type
         description: name
         value
@@ -93,6 +94,7 @@ export class DgraphDataSource extends RESTDataSource {
     return `{
       query (func: eq(dgraph.type, "Product")) @cascade {
         code
+        uid
         type: dgraph.type
         description: name
         properties {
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Loooks good, Do you have an example query that was causing this error for testing?

lache-melvin:
> Do you have an example query that was causing this error for testing?

I was actually getting it when I hadn't re-uploaded the DGraph GraphQL schema, so I think the whole endpoint was returning an empty `data` object. Seems like a hole we should fix in general... though now I'm thinking about it, it potentially could be worth returning an error if we don't get the expected response body from DGraph, rather than just returning an empty array...

jmbrunskill:
> though now I'm thinking about it, it potentially could be worth returning an error if we don't get the expected response body from DGraph, rather than just returning an empty array...

Yeah if it's an error we should probably return an error In that sort of case think.

jmbrunskill:
```query MyQuery {
  entity(code: "12e6911d") {
    type
    code
    uid
    description
    interactions {
      name
      description
      severity
      rxcui
      source
    }
    parents {
      code
    }
    product {
      code
    }
    properties {
      type
      value
    }
  }
}```

Similar error?

```
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field EntityType.uid.",
      "locations": [
        {
          "line": 5,
          "column": 5
        }
      ],
      "path": [
        "entity",
        "uid"
      ],
.
.
.
```

lache-melvin:
Returning the failed DGraph query error: [84d40ae](https://github.com/msupply-foundation/unified-codes/pull/428/commits/84d40aee7c0570f0ad24cd77ae2a613db74a0d6a)

Including UIDs: [3a4a195](https://github.com/msupply-foundation/unified-codes/pull/428/commits/3a4a19558b0bf24cfcdc206f8d1a875b6ef867d1)

PR #434, opened 2023-12-11
431 new frontend browse

lache-melvin:
First half of #431 

## Description

- Adds `@uc-frontend/system` package
- Adds the browse view to the new frontend
- Implements search and filter by drug/consumable, as was available in the old UI
  - You can sort by description, ascending or descending
  - There are other filter parameters in the `entities` query, but I don't think they're handled in the backend yet... 
- Removes the left hand side nav drawer (to be brought back only for admins...)

Would be good to get thoughts on putting a max width on the table? In all the implementations we've had so far our tables stretch to full width, but especially without the nav drawer, this starts to look weird on larger screens

Screenshot 2023-12-12 at 7 34 42 AM


frontend/system/src/Entities/ListView.tsx
```diff
@@ -0,0 +1,121 @@
...
+      categories,
+      description: search,
+      orderBy: {
+        field: sortBy.key,
+        descending: sortBy.isDesc,
+      },
+    },
+    first,
+    offset,
+  });
```
lache-melvin, line 58: The `useQueryParams` hook we provide from our `common` package returns are very specific structure, targeted for the filters in our Rust backend (and the `useColumns`/`DataTable` components in the frontend). This backend is a little different, so instead of being able to simply do `useEntities(queryParams)`, I have to do the above mapping (finding the `searchFilter` piece particularly unappealing.)

I thought about implementing a new `useQueryParams` hook specific for this repo? But then wasn't sure how set we are on this different structure from the backend perspective, particular given half of the fields on `filter` aren't even being handled yet...

frontend/system/src/Entities/ListView.tsx
```diff
@@ -0,0 +1,121 @@
...
+      { key: 'type', label: 'label.type', sortable: false, width: 200 },
+    ],
+    { sortBy: sortBy, onChangeSortBy: updateSortQuery },
+    [sortBy, updateSortQuery]
+  );
+
+  const [categories, setCategories] = useState<string[]>(['drug']);
+
+  const searchFilter = filter.filterBy?.['search'];
+  const search = typeof searchFilter === 'string' ? searchFilter : '';
```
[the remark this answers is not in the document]
jmbrunskill: I think it you add `{match:'contains'}` to your graphql query you'd get that behaviour by default?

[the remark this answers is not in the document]
jmbrunskill: Adding a `*` might be a good temporary solution though.

frontend/system/src/Entities/ListView.tsx
```diff
@@ -0,0 +1,121 @@
...
+      orderBy: {
+        field: sortBy.key,
+        descending: sortBy.isDesc,
+      },
+    },
+    first,
+    offset,
+  });
+
+  const updateCategories = (category: string) => {
```
jmbrunskill, line 60:
```suggestion
  const toggleCategory = (category: string) => {
```
Not a major, just had to check the code to understand what it did

frontend/system/src/Entities/ListView.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <TableProvider createStore={createTableStore}>
+        <AppBarContentPortal
+          sx={{
+            paddingBottom: '16px',
+            flex: 1,
+            display: 'flex',
+            justifyContent: 'space-between',
+          }}
+        >
+          <SearchToolbar filter={filter} />
```
jmbrunskill, line 88: This search control is still experiencing that weird behaviour where we click the `x` button and the box just get smaller. I thought it was fixed in notify but maybe not? Can you check?

jmbrunskill (approved):
Looks good, great base to iterate on.

Side Note:
Might be worth investigating this 1 number difference, probably not PR related though.

image

image

lache-melvin:
> Looks good, great base to iterate on.
> 
> Side Note: Might be worth investigating this 1 number difference, probably not PR related though.
> 
> image
> 
> image

Oh interesting... my local version of the old UI has 452 as well - am I missing a drug? Do I have old CSV? Is the 453 screenshot from your local instance or the deployed one?

frontend/system/src/Entities/ListView.tsx
lache-melvin: Will add a ticket to circle back on this 👍

frontend/system/src/Entities/ListView.tsx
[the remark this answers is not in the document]
lache-melvin: Oh nice! Better with `{ match: 'contains' }` for sure! [99508ba](https://github.com/msupply-foundation/unified-codes/pull/434/commits/99508ba9843bbe3b98b8e7117af4a50448b23e6e)

frontend/system/src/Entities/ListView.tsx
lache-melvin: Yeah nice catch, [0c0918d](https://github.com/msupply-foundation/unified-codes/pull/434/commits/0c0918d9debe2a13a159897173927a66934a0c1d)

frontend/system/src/Entities/ListView.tsx
lache-melvin: Thanks for spotting this! Turns out it was... half fixed in Notify, but there was some kind of race condition between the onClick firing as expected, or the focus shifting away from input before it could get there. I've [updated it to use `onMouseDown`](https://github.com/msupply-foundation/unified-codes/pull/434/commits/f0ef8b2782063bb30a1cc41979117c85c54199db) instead, which appears to be working reliably! Will make a PR for Notify (and HSH if the same thing is there?) now!

PR #443, opened 2023-12-12
435 poc dgraph from rust

### LINKED ISSUE #435: POC - Dgraph From Rust!
## Expected Behaviour
POC to see if it would be easy to query dgraph from a rust backend...

jmbrunskill:
Closes #435

## Description
Proof of concept querying dgraph from a rust backend.

The `v1` API is a separate package to the main graphql endpoint.
This means there's good separation between admin functions and non-authenticated public api.

The backed is based on notify, with some things pulled out and cleaned up.
Emails have stayed in, as they can enable things like password reset which would be useful.

At this stage it only implements the `entity` v1 query, which allows lookup of a universal code by it's code.

### Tests:
At the moment, this only passes two integration tests for the `entity` request.
`entities` and filter will come in a later PR.

`v1_graphql_entity.test.ts` is currently setup to point to port 4007 (rust port)

Run using `yarn test-integration`

### Notes:

I haven't implemented Drug Interactions, not sure if we need to or not?

I have only gone up one level for parents in the query. Wondering if I should just query all the parents and children in dGraph? Seems really fast? Feels wasteful allocating all the memory etc when most queries won't need it.
The other option is to do a loader...

I did the single parent thing to allow matching the logic here, to return `drug` or `consumable` based on it's parent node: https://github.com/msupply-foundation/unified-codes/blob/276158a1aa13ec911610ff3759ae7cf8b527d19c/data-service/src/v1/types/DgraphDataSource.ts#L200

> `entities` will come in a separate PR after this one...


data-service/src/integration/v1_graphql_entity.test.ts
```diff
@@ -7,7 +7,8 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const crossFetch = require('cross-fetch');
 
-const url = 'http://localhost:4000/v1/graphql';
+// const url = 'http://localhost:4000/v1/graphql';
+const url = 'http://localhost:4007/v1/graphql';
```
jmbrunskill, line 11: image

Any idea how to resolve this error?

data-service/src/v1/dataservice-schema.graphql
```diff
@@ -0,0 +1,90 @@
+type Query {
```
jmbrunskill, line 1: This isn't actually required, but keep it here for now as reference for the old API

backend/graphql_v1/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "graphql_v1"
```
jmbrunskill, line 2: This package is new, might be worth a closer look than other packages?

backend/service/src/service_provider.rs
```diff
@@ -0,0 +1,81 @@
...
+    log_service::{LogService, LogServiceTrait},
+    settings::Settings,
+    universal_codes::UniversalCodesService,
+    user_account::{UserAccountService, UserAccountServiceTrait},
+};
+
+pub struct ServiceProvider {
+    pub connection_manager: StorageConnectionManager,
+    pub email_service: Box<dyn EmailServiceTrait>,
+    pub dgraph_service: Box<UniversalCodesService>,
```
jmbrunskill, line 17: Didn't feel worth creating a Trait for this, especially as it needs to be async and that requires a special async trait library. I think we can assume that the test cases can run against a real dgraph database locally with all the universal codes data. In other words, I don't need to mock this...

data-service/src/integration/v1_graphql.entities.test.ts
```diff
@@ -0,0 +1,400 @@
+/**
+ * GraphQL API Integration tests
+ *
+ * @group integration
+ */
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const crossFetch = require('cross-fetch');
+
+const url = 'http://localhost:4000/v1/graphql';
```
jmbrunskill, line 10: These are still looking at the old API at the moment, until I re-implement in Rust

data-service/src/integration/v1_graphql.entities.test.ts
```diff
@@ -0,0 +1,400 @@
...
+    }),
+  })
+    .then(res => res.json())
+    .then(res => {
+      expect(res).toEqual(expected);
+    });
+});
+
+// Note: This test fails as the old API doesn't implement code search!!!
+test('filter-code-exact', () => {
```
jmbrunskill, line 312: Yup this does fail with old api!

jmbrunskill:
@lache-melvin Sorry, I should have done this as 2 PRs, one for the backend base, and another for the new graphql_1 stuff.
I've tried to highlight the main places to look with comments.

data-service/src/integration/v1_graphql_entity.test.ts
mark-prins: the `crossFetch`? can't you just use a different name for the variable?
`const cf = require('cross-fetch');`

though it would also be preferable to use `import` rather than `require`

`import cf from 'cross-fetch';`

backend/server/src/serve_frontend.rs
```diff
@@ -0,0 +1,68 @@
...
+    http::{header::ContentType, StatusCode},
+    web::ServiceConfig,
+    HttpRequest, HttpResponse, Responder,
+};
+use mime_guess::{from_path, mime};
+use rust_embed::RustEmbed;
+
+#[derive(RustEmbed)]
+// Relative to server/Cargo.toml
+#[folder = "../../frontend/packages/host/dist/"]
```
lache-melvin, line 12:
```suggestion
#[folder = "../../frontend/host/dist/"]
```

`packages` dir was removed as part of the typescriptification... if we're going to our rust backend perhaps one could argue for the return of it though!

backend/configuration/example.yaml
```diff
@@ -0,0 +1,17 @@
+# To use custom server settings for a local or a production environment create a copy
+# of this example file and name it local.yaml
+server:
+  host: 127.0.0.1
+  app_url: 'http://localhost:3007'
+dgraph:
+  host: 'localhost'
```
lache-melvin, line 7:
```suggestion
  host: 'http://localhost'
```

My tests were failing from this...

lache-melvin (approved):
Couple of little things I picked up but looks awesome James! 💪

data-service/src/integration/v1_graphql_entity.test.ts
jmbrunskill: I'll put the `export {}` thing in for now :)

backend/configuration/example.yaml
```diff
@@ -0,0 +1,17 @@
+# To use custom server settings for a local or a production environment create a copy
+# of this example file and name it local.yaml
+server:
+  host: 127.0.0.1
+  app_url: 'http://localhost:3007'
```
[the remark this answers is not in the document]
jmbrunskill: Yeah I found this too, thanks!

backend/graphql_v1/core/src/standard_graphql_error.rs
```diff
@@ -0,0 +1,35 @@
...
+#[derive(Debug, Error, Clone)]
+pub enum StandardGraphqlError {
+    #[error("Internal error")]
+    InternalError(String),
+
+    #[error("Bad user input")]
+    BadUserInput(String),
+
+    #[error("Unauthenticated")]
+    Unauthenticated(String),
```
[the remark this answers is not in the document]
jmbrunskill: Yup good point

backend/configuration/example.yaml
jmbrunskill: 3f9cd65

backend/server/src/serve_frontend.rs
jmbrunskill: 3f9cd65

backend/graphql_v1/core/src/standard_graphql_error.rs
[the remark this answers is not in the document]
jmbrunskill: 3f9cd65

backend/configuration/example.yaml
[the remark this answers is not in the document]
jmbrunskill: 3f9cd65

backend/server/debian/notify.service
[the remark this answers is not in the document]
jmbrunskill: 3f9cd65

PR #449, opened 2023-12-15
442 new medicine UI

### LINKED ISSUE #442: Create UI for adding a new drug

lache-melvin:
Fixes #442

## Description

Adds a new Admin service to the frontend (i.e. `/admin`)

Adds a form for adding new drugs. Resulting state is just being console.logged for now, next step will be persisting it!

Navigate to `/admin/new-drug`:
Screenshot 2023-12-15 at 1 55 18 PM

Adding properties:
Screenshot 2023-12-15 at 1 44 40 PM

Dropdowns for some categories:
Screenshot 2023-12-15 at 1 54 39 PM

Future things to do:
- Persisting the result
- Validation of units
- Separate form for consumables


frontend/system/src/Admin/DrugEditForm/AddFieldButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <FlatButton
+      startIcon={<PlusCircleIcon />}
+      label={label}
+      onClick={onClick}
+      disableFocusRipple
+      sx={{
+        marginLeft: '20px',
+        '&.Mui-focusVisible': {
+          backgroundColor: '#e95c3029',
```
lache-melvin, line 21: The FlatButton has a strange bug where the focus ripple animation doesn't stop when you hit enter on the button, which takes the focus elsewhere on the page. 

Disabling the focus ripple and adding a background colour when the button is in focus appears to work reliably for both mouse and keyboard interactions

frontend/system/src/Admin/DrugEditForm/categories.ts
```diff
@@ -0,0 +1,85 @@
...
+    { label: 'Solution (Intramuscular)', value: 'Solution (Intramuscular)' },
+    { label: 'Solution (Subcutaneous)', value: 'Solution (Subcutaneous)' },
+    { label: 'Spray', value: 'Spray' },
+    { label: 'Suppository', value: 'Suppository' },
+    { label: 'Suspension', value: 'Suspension' },
+    { label: 'Tablet', value: 'Tablet' },
+    { label: 'Tablet (Chewable)', value: 'Tablet (Chewable)' },
+    { label: 'Tablet (Crushable)', value: 'Tablet (Crushable)' },
+    { label: 'Tablet (Dispersible)', value: 'Tablet (Dispersible)' },
+    { label: 'Tablet (Enteric-coated)', value: 'Tablet (Enteric-coated)' },
```
lache-melvin, line 56: Have added some dose form qualifications here, as asked by Louisa to have them merged in with the dose form options, but will need to have her/someone go through to determine which are valid dose form qualifications for particular dose forms!

jmbrunskill (approved):
Looks good!
The main thing I'm not sure about it, is there's no way to skip a level in the hierarchy.

Seems we are using that for some things like `https://codes.msupply.foundation/detail/7786789` (No strength)

Doesn't display very nicely anyway, but maybe we want to maintain the ability to do this?

jmbrunskill:
One other UX thought, it feels more natural for me when I want add a new form or something to want to go back to the heading and click and add button there. Scrolling down though a big list feels a bit weird, but we'll see how the feedback is I guess...

frontend/system/src/Admin/DrugEditForm/categories.ts
lache-melvin: I think the custom comments/details sounds like the way to go, as long as that resonates with Louisa too... referring to that SOP Rowena wrote, it looks like it was at least intended that there would only be a predefined set of dose forms and qualifications, with additions to be reviewed first. I'm assuming those additional form qualifications for the vaccines came about because there was nowhere better to store that information? Though we'll need to think about what we do with the existing data that looks like that if we go the comments route...

And yep managing the options in the backend for sure!

lache-melvin: Created #452 for this!

lache-melvin:
> One other UX thought, it feels more natural for me when I want add a new form or something to want to go back to the heading and click and add button there. Scrolling down though a big list feels a bit weird, but we'll see how the feedback is I guess...

Ah interesting! Yeah this would be a great thing to get user feedback on, especially with the bigger lists!

lache-melvin:
> Looks good! The main thing I'm not sure about it, is there's no way to skip a level in the hierarchy ... maybe we want to maintain the ability to do this?
  
Ah yeah, I think you're right. I started thinking about making this change, but I think it would be worth getting some other input first...

I had a look at the SOP, and according to that, after the product name all the levels are optional, except that one of strength or unit must be provided. I assume this if for the mSupply query, but it's an interesting constraint to enforce in this form structure... perhaps the table version will be needed?

Also, the SOP doesn't speak to which levels are able to be specified if the parent wasn't provided, and given all the drug data so far at least has routes and forms, and that the SOP is a few years old, I'm wondering if its worth another discussion around what we constrain and allow.

I'll pop a message in our group chat, and create an issue to come back to this!

frontend/system/src/Admin/DrugEditForm/PropertiesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+              onClose();
+            }}
+            isLoading={false}
+            startIcon={<ArrowRightIcon />}
+            sx={{ marginLeft: 1 }}
+          >
+            {t('label.add-properties')}
+          </LoadingButton>
+        }
+        cancelButton={<DialogButton variant="cancel" onClick={onClose} />}
```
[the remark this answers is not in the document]
lache-melvin: Oo great catch! [a6f9649](https://github.com/msupply-foundation/unified-codes/pull/449/commits/a6f96497bab5e09f92f07192475c407cf5839a73)

frontend/system/src/Admin/DrugEditForm/PropertiesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+          <LoadingButton
+            onClick={() => {
+              onSave(properties);
+              onClose();
+            }}
+            isLoading={false}
+            startIcon={<ArrowRightIcon />}
+            sx={{ marginLeft: 1 }}
+          >
+            {t('label.add-properties')}
```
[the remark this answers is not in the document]
lache-melvin: [a43b725](https://github.com/msupply-foundation/unified-codes/pull/449/commits/a43b725f8d152dbbe2c0d93d32eb3644b64b689f)

frontend/system/src/Admin/DrugEditForm/PropertiesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  const columns = useColumns<Property>([
+    {
+      key: 'type',
+      width: '50%',
+      label: 'label.type',
+      sortable: false,
+      Cell: ({ rowData, rows }) => (
+        <Select
```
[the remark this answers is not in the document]
lache-melvin: [6c79914](https://github.com/msupply-foundation/unified-codes/pull/449/commits/6c799149692ce20f3570bbbdbadc9d53308f980d)

PR #454, opened 2023-12-18
450 add admin login

### LINKED ISSUE #450: Add admin login
- Put login around the Admin service (i.e. `/admin`)
- Bring back the side nav with the logout button
- To figure out - showing the browse/details within the admin view?

lache-melvin:
Fixes #450 

## Description

Navigate to `/login` to login as an admin!

- Adds login for admin
- Adds user account management back in
   - This includes adding and inviting users, forgot password and password reset
   - Currently this doesn't include logs on the user, but easily could? Would that info be useful in this app?
- Wraps the user management and the new drug adding to require authentication
- Adds AppDrawer back in for the admin to navigate
  - This does mean that the logo man is a 'click to return home' icon when unauthenticated, and a 'click to open/close the nav drawer' when authenticated... how much do we dislike this?
  

## Note to self: THIS PR IS READY TO MERGE, JUST WAITING FOR THE ENTITIES QUERIES TO BE AVAILABLE ON `/GRAPHQL`
This can't be merged until the entity queries are available at `/graphql`... the browse/detail pages are just returning errors for now. I'll also need to rerun the codegen before merging!


frontend/system/src/Admin/Users/ListView/UserAccountEditModal/UserAccountEditModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const { draft, password, setPassword, onUpdate, onSave, isLoading } =
+    useDraftUserAccount(userAccount, mode);
+
+  const isInvalid =
+    (!isValidUsername(draft.username) ||
+      (mode === ModalMode.Create && !password) ||
+      draft.permissions.length === 0) ??
+    true;
+
+  // TODO: do we want the logs tab here too?
```
jmbrunskill, line 113: I think an audit log is probably going to be useful, but I'd like to do it more like the events in notify, so it can be searched in one place, and we don't need to do loaders on all our user account objects?

frontend/system/src/Admin/Users/api/operations.graphql
```diff
@@ -0,0 +1,74 @@
+fragment UserAccountRow on UserAccountNode {
+  __typename
+  id
+  username
+  email
+  displayName
+  permissions
+  auditLogs {
```
jmbrunskill, line 8: Can take this out if we're not displaying them?

jmbrunskill (approved):
🚀
Looks great!

frontend/system/src/Admin/Users/api/operations.graphql
lache-melvin: Great catch thanks [823a790](https://github.com/msupply-foundation/unified-codes/pull/454/commits/823a790facafe553f9c76ae3e681f36825ca06b4)

frontend/system/src/Admin/Users/ListView/UserAccountEditModal/UserAccountEditModal.tsx
lache-melvin: Yeah nice, I did like that in Notify, that should work well :)

PR #462, opened 2023-12-19
455 configuration edit UI

### LINKED ISSUE #455: Add UI for editing category config
i.e. the different drug forms, routes etc.  

These should be a predefined dropdown list in the drug edit form, but from a different screen we want to be able to add/edit/remove options! 

Also to be considered - when these options are selected, should we store a reference to them on the drug (e.g. so if a typo is fixed in this new page, that gets dispersed to all the drugs? Or just copy the string, so that we can edit these options freely without worrying to support old ones.

lache-melvin:
Fixes #455

## Description

- Adds UI for managing config
- Tabs for managing categories (i.e. Routes/Forms)
- Tab for managing properties
  - Have not implemented delete for properties, as these are used as reference data for other code... but could implement a deactivate or something in future if it is desired
  - Is the helper text in the properties add/edit modal ok?

Screenshot 2023-12-20 at 12 34 45 PM
Screenshot 2023-12-20 at 12 34 54 PM
Screenshot 2023-12-20 at 12 35 08 PM


jmbrunskill:
> * Is the helper text in the properties add/edit modal ok?

Yup looks good. I'm slightly worried about using {{code}} for non universal codes. Seems slightly confusing to me, but let's check with others...

frontend/system/src/Admin/Configuration/CategoryConfigTab.tsx
```diff
@@ -0,0 +1,114 @@
...
+  label: string;
+  value: string;
+};
+
+type CategoryConfigTabProps = {
+  data: CategoryOption[];
+  category: string;
+};
+
+const CategoryConfigTabComponent = ({
```
jmbrunskill, line 35: The name `CategoryConfigTab` confused me a little bit, as we have categories for `Drug` and `Consumable`
Maybe `ListConfigTab`? Can't think of a perfect name :)

jmbrunskill: The `Categories` are `types` in the dgraph but that's also confusing.
If we're considering changing a lot about the dgraph, we could make category the thing like `drug`, `unit` etc?
Maybe that's another approach...

jmbrunskill (approved):
Looks good, will be good to try it out properly once saving is implemented.

jmbrunskill:
It's not a major as it won't be used a lot, but the delete feels very disconnected from the selection box...
image

lache-melvin:
> Yup looks good. I'm slightly worried about using {{code}} for non universal codes. Seems slightly confusing to me, but let's check with others...

Ah yeah good point, we can ask around tomorrow?

frontend/system/src/Admin/Configuration/CategoryConfigTab.tsx
lache-melvin: Gone with `OptionList`... not sure I love that name either but at least I don't think its in conflict with anything else 😅  [a0c6cbe](https://github.com/msupply-foundation/unified-codes/pull/462/commits/a0c6cbe20854227bf55890d6b876cde11fd1e1c8)

lache-melvin:
> It's not a major as it won't be used a lot, but the delete feels very disconnected from the selection box... image

Yeah... I tried putting it within that main grey box but it looked a bit strange, pushing the rest of the content down... keen for any ideas - would also be great to get Anisha's eye over the app in general at some point in the new year!

PR #464, opened 2023-12-19
fix browse query

### LINKED ISSUE #463: Fix frontend browse query
Old API filtered for products, but new one doesn't so our browse page now gets all the levels... include a `type: 'drug'` filter to get the same result!

Image

lache-melvin:
Fixes #463

## Description

Adds the `type: drug` filter so we get the same result as we used to!


jmbrunskill (approved):
⛴ it!

PR #465, opened 2023-12-19
Add upsert_entity function

jmbrunskill:
Fixes: #461

Really hacky solution for updating entities in dgraph


backend/dgraph/src/upsert_entity.rs
```diff
@@ -0,0 +1,179 @@
...
+#[allow(non_snake_case)]
+#[derive(Deserialize, Debug, Clone)]
+pub struct UpsertResponseData {
+    pub data: UpsertResponse,
+}
+
+#[allow(non_snake_case)]
+#[derive(Deserialize, Debug, Clone)]
+pub struct UpsertResponse {
+    pub numUids: u32,
```
jmbrunskill, line 21: Since we don't know what type of mutation we're going to do, we can't return the entity data as part of the query :(

data-loader/data/v2/schema.graphql
```diff
@@ -4,7 +4,7 @@
 
 interface Entity {
   id: ID!
-  code: String @dgraph(pred: "code") @search(by: [exact, trigram])
+  code: String! @dgraph(pred: "code") @id @search(by: [exact, trigram])
```
jmbrunskill, line 7: BTW: This allows us to update via code (Rather than the internal ID)

backend/service/src/universal_codes/upsert.rs
```diff
@@ -0,0 +1,158 @@
...
+pub async fn validate(
+    // ctx: &ServiceContext,
+    client: &dgraph::DgraphClient,
+    new_entity: &UpsertUniversalCode,
+) -> Result<Option<Entity>, ModifyUniversalCodeError> {
+    // Check if entity exists with the same description
+    if let Some(description) = new_entity.description.clone() {
+        let duplicated =
+            check_description_duplication(client, description, new_entity.code.clone()).await?;
+        if duplicated.is_some() {
```
lache-melvin, line 125: The upsert test is failing here for me? `duplicated` is always `Some`, but the data array inside should be empty when not a duplicate...

lache-melvin (approved):
Found a couple of little things that are breaking for me but on the whole this looks great James!! 

Would be awesome if we can get the logging going, but I'd also be down to get this merged in whatever capacity, so we can get _something_ saving, and come back to it 👍

backend/service/src/universal_codes/upsert.rs
jmbrunskill: Thanks good find!

frontend/config/src/index.ts
```diff
@@ -16,7 +16,7 @@ declare global {
 }
 
 const {
-  API_HOST = 'http://localhost:4000/v1',
+  API_HOST = 'http://localhost:4000',
```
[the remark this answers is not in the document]
jmbrunskill: Updated from Develop!

backend/service/src/universal_codes/upsert.rs
jmbrunskill: Fixed now (I think)

PR #467, opened 2023-12-20
466 edit drug UI

### LINKED ISSUE #466: Edit existing drug UI
Edit is a bit incorrect - we allow adding new strengths/routes etc., but not editing existing data

lache-melvin:
Fixes #466

## Description

Adds an update button to the entity details page when logged in as an admin:
Screenshot 2023-12-21 at 10 52 04 AM

Populates the drug edit form with existing entity data. New fields can be added, but existing fields are disabled:
Screenshot 2023-12-21 at 10 52 27 AM

Have left it so existing properties can be edited - I think this is fine, it's only the universal code data we don't want to change...

Removed the immediate packaging section of the edit drug form - would be good to get other thoughts on this, as we are capturing immediatePackaging/packSize in the spreadsheet, but the UI hasn't been querying for it, and neither does the mSupply query. Wouldn't be great to enter that data and then not see it displayed, but not sure if the solution is to not capture that extra data, or to update the UI to display it!


lache-melvin:
Also not implemented is building the object we'll send to the backend to save... this should probably come once the mutation structure/pattern has been defined - not sure whether we would send the full updated entity and determine the changes on the backend, or only send the new nodes.

jmbrunskill:
One small usability thing, it would be nice to be able to remove something before you save it at least.

For example if I accidentally add two unit's but actually only want one...
image

It's not clear right now what I should do e.g.
- Will the extra blank unit get created?
- Will I get a validation error that I can't correct?

It's even worse for `form` as there' no way to select nothing...

frontend/system/src/Admin/EditDrug/helpers.test.ts
```diff
@@ -0,0 +1,256 @@
...
+              code: '66e85500',
+              name: 'Solution',
+              type: 'form', // include: direct child of form_category (route)
+              properties: [],
+              children: [],
+            },
+            {
+              code: 'e4edcb00',
+              name: '15mL',
+              type: 'unit_of_use', // exclude: not a direct child of route
```
jmbrunskill, line 224: Do you think it's plausable/easy to skip the levels and add a blank editable node in between?
Then for fixing issues like this we can use the UI rather than more spreadsheet hacking...

jmbrunskill (approved):
Nice this is great work!

jmbrunskill:
> * Will the extra blank unit get created?
> * Will I get a validation error that I can't correct?

Tried to improve this issue by adding delete buttons on editable nodes 
[48fc383](https://github.com/msupply-foundation/unified-codes/pull/467/commits/48fc383f5165a7007629cb678840fb85047ea524)

image

frontend/system/src/Admin/EditDrug/helpers.test.ts
lache-melvin: As we chatted about this morning I think that definitely could be an option. Maybe we hold fire on implementing this until we chat with Craig? If we're only implementing it to resolve issues from the current spreadsheet, we should probably just fix them in the spreadsheet. Wouldn't want to be opening ourselves up to missing data if that's not what we're wanting?

However if we do need to support skipping levels, I think we could just remove some of the validation and we'll be good to go 🚀  (Though I would have some questions, like does a blank node still get a code?)

PR #481, opened 2023-12-21
479 pending changes list view UI

### LINKED ISSUE #479: Create UI for pending changes list

lache-melvin:
Fixes #479

## Description

Adds the list view for pending changes, with dummy data:
Screenshot 2023-12-22 at 10 37 37 AM


frontend/system/src/Admin/PendingChanges/ListView.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+// TODO: types from codegen
+type PendingChange = {
+  id: string;
+  productName: string;
+  category: string;
+  requestDate: string;
+  changeType: 'Addition' | 'New';
+  requestedBy: string;
+  for: string;
```
jmbrunskill, line 23: I guess for this we'll need some mechanism to record this? Perhaps when saving the the entity changes?

jmbrunskill (approved):
Looks good, but as an aside I think we should be putting a bit more detail in our Issue so we know what's expected in the PR. For example we'll need to thinking about how to store the data and detect states such as `New` vs `Addition` and Who requested it and how.

lache-melvin:
> Looks good, but as an aside I think we should be putting a bit more detail in our Issue so we know what's expected in the PR. For example we'll need to thinking about how to store the data and detect states such as `New` vs `Addition` and Who requested it and how.

Yup fair call thank you! I was getting pretty lazy just converting draft todos to issues without fleshing them out 🙈

frontend/system/src/Admin/PendingChanges/ListView.tsx
lache-melvin: Yeah... my thinking was that this slots in the middle somewhat? We can save a PendingChange when a drug is added/updated, and then run the upsert entity when this is approved?

lache-melvin: OH haha I see what you're saying, how are we capturing who the change is for... yeah I guess at the same time of saving the add/edit... perhaps we could have a modal popup on save the entity changes that asks who the change is for?

jmbrunskill: Let's discuss with Jess/Louisa what they want to see when approving. Was assuming it was just the name of the person who entered it.

jmbrunskill: > Yeah... my thinking was that this slots in the middle somewhat? We can save a PendingChange when a drug is added/updated, and then run the upsert entity when this is approved?

I like this approach, will have a think about it. Could have some kind of PendingEntityType in dGraph and return them both or something....

lache-melvin: > Let's discuss with Jess/Louisa what they want to see when approving. Was assuming it was just the name of the person who entered it.

Oh I think in the mock ups Louisa did, `requestedBy` was the person that entered it and `for` was the country they were making the change for. If it's a little clunky for us to capture that in the UI we could find out actually how useful that info would be for them though?

PR #485, opened 2023-12-22
484 update setup instructions

### LINKED ISSUE #484: setup instructions are out of date
we upgraded to a rust backend!

lache-melvin:
Fixes #484 

## Description

Update the readme setup instructions, as well as adding some extra yarn scripts, after we've brought in the Rust backend.

This is a slightly different approach than usual, where we keep the frontend and backend dirs and the running of their scripts separate... it may (?) be worth going back to the usual pattern, but this would require another folder rearrange. 

Also puts the dgraph tests behind a feature flag so that the backend tests can run in CI!


README.md
```diff
@@ -16,32 +16,30 @@ The test framework is jest; functional areas are separated into packages and man
...
 - Initialise DGraph schema and load data into DGraph:
   - `yarn load-data`
-- Start the data-service:
-  - `yarn start-data-service`
-  - Access v1 api at http://localhost:4000/v1/graphql
-  - Access v2 api at http://localhost:4000/v2/graphql
+- Start the backend:
+  - `yarn start-backend`
+  - Access v1 api at http://localhost:4007/v1/graphql
+  - Access v2 api at http://localhost:4007/graphql
```
jmbrunskill, line 22:
```suggestion
  - Access internal api at http://localhost:4007/graphql
```

I'm thinking of the `/graphql` endpoint as something we can iterate on and use in the frontend, not a published API for external users...

backend/dgraph/Cargo.toml
```diff
@@ -7,6 +7,9 @@ edition = "2018"
 doctest = false
 
 [dependencies]
-serde = {version = "1.0.126", features = ["derive"]}
-tokio = {version = "1", features = ["full"]}
+serde = { version = "1.0.126", features = ["derive"] }
+tokio = { version = "1", features = ["full"] }
 gql_client = "1.0.7"
+
+[features]
```
jmbrunskill, line 14:
```suggestion
[features]
default = ["dgraph-tests"]
```

.github/workflows/CI.yml
```diff
@@ -10,18 +10,11 @@ jobs:
...
         run: yarn build-frontend
 
       - name: Test frontend
         run: yarn test-frontend
 
-      # This only runs the unit tests - integration tests in CI would be more expensive
-      - name: Test data-service
-        run: yarn test-data-service
+      - name: Test backend
+        run: yarn test-backend
```
jmbrunskill, line 20:
```suggestion
        run: yarn test-backend --no-default-features
```

jmbrunskill: Not actually sure if this will work, might need to do something special to pass the parameters through yarn?

backend/dgraph/Cargo.toml
jmbrunskill: This is basically because I feel like it would be good to run our dgraph tests regularly as devs, so would rather disable them in CI rather than have to enable the every time in our IDE.
Although we can set features on for a VS code workspace so I'm happy if you leave it as is.

jmbrunskill (approved):
I like everything in one place like this.
Nice touch with the `dgraph-tests` feature, I was actually thinking about doing something like this today.

backend/dgraph/Cargo.toml
lache-melvin: Nah nice I like this!

.github/workflows/CI.yml
lache-melvin: Ok never mind... yarn passes through the parameters just fine, but --no-default-features appears to only work if you define a target package.

i.e. 
- `cargo test --no-default-features --package dgraph` works (resulting in no tests running, all tests in dgraph package are behind the flag) 
- However `cargo test --all --no-default-features` doesn't work, dgraph tests still run 
- Explicitly defining all the packages also doesn't work... `cargo test --no-default-features --package "dgraph service"`

The internet doesn't seem to have an easy fix, except perhaps virtual workspaces... have you used/done this before or shall I just keep it as is?

jmbrunskill: I had a quick look but agree it doesn't seem to work.
I think it's because the dependancies are supposed to specify if they want to enable default features or not. Seems like that's not possible in workspace members.


https://github.com/rust-lang/cargo/pull/11409
This PR fixes it for workspace dependencies but not for members. Let's leave it as it was.

PR #486, opened 2023-12-22
477 fix links and redirects

### LINKED ISSUE #477: Admin breadcrumbs take you to blank pages...
If you are on say `/admin/edit/c2v3b4b`, the App Bar displays breadcrumbs: `Admin / Edit / {drugName}`. The `Admin` and `Edit` are clickable, and would navigate the user to `/admin` or `/admin/edit` respectively.

Often-times that would be great, but we at least currently don't have any page on the `/admin` or `/admin/edit` routes! So the user sees a blank square. 

We should either prevent those breadcrumbs from being clickable, or at least put some kind of wildcard on the admin router, so if the route doesn't exist, the user gets navigated back to the browse page!

lache-melvin:
Fixes #477

## Description

Redirects users away from `/admin/{something-unknown}` paths back to `/browse`, rather than showing the 404 page. This is due to the fact that the breadcrumbs are clickable, but don't actually direct the user to an existent page. Clicking a link within the app and getting taken to a 404 always has the feel of something being broken, so this seemed like the easiest next best thing.

Also moves the Paper/Box wrapper so the 404 page doesn't render within it - that looked weird... this is the before:

Screenshot 2023-12-22 at 3 57 39 PM

### DIFF
--- a/frontend/host/src/Site.tsx
+++ b/frontend/host/src/Site.tsx
@@ -15,6 +15,7 @@ import {
   Navigate,
   Paper,
   useAuthContext,
+  PropsWithChildrenOnly,
 } from '@uc-frontend/common';
 import { AppBar, AppDrawer, Footer, NotFound } from './components';
 import { CommandK } from './CommandK';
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
I would be ok with a 404 page, but this seems good too.

PR #489, opened 2024-01-03
488 save a new drug directly to dgraph db

### LINKED ISSUE #488: Save a new drug directly to dgraph db
## Expected Behaviour
The save button in the UI should work!

## Current Behaviour
Saving doesn't do anything...

## Possible Solution

- [ ] ID Should be created by the backend (UUID is ok for now, but should be something more consistent)

jmbrunskill:
Fixes #488 

## Description

Recursively adds/updates entity nodes sent from the frontend.
Any new properties are added too if they're new but they're not updated. (To fix this we'd need to call the updateProperty graphql request in the dgraph - new issue to be created, but only if we want to allow updates...)


frontend/system/src/Admin/EditDrug/helpers.ts
```diff
@@ -55,11 +56,78 @@ export const buildDrugInputFromEntity = (entity: EntityDetails): DrugInput => {
...
+export const buildEntityFromDrugInput = (
+  drug: DrugInput
+): UpsertEntityInput => {
+  return {
+    parentCode: '933f3f00', // Drug parent code
+    code: drug.code,
+    name: drug.name,
+    description: drug.name,
+    type: 'Product',
+    category: 'Drug',
```
lache-melvin, line 68: NIT for now, but wondering about defining these types and categories as enums on the backend that we could expose for the frontend? Save us some inevitable typo?

lache-melvin (approved):
Looks awesome James! Works well for me 🚀

lache-melvin:
> Any new properties are added too if they're new but they're not updated. (To fix this we'd need to call the updateProperty graphql request in the dgraph - new issue to be created, but only if we want to allow updates...)

This makes sense - might be good to disable those existing property fields until we allow updates (if we do)?

frontend/system/src/Admin/EditDrug/helpers.ts
jmbrunskill: Yeah I think this is a good idea. However, the backend doesn't know about this enums either (I don't think) they're defined in the loader at the moment and stored as strings... So maybe frontend enum/strings are just as good?

lache-melvin: Yeah that would be good too! 🚀

jmbrunskill: https://github.com/msupply-foundation/unified-codes/pull/489/commits/5d5d07e210ab6cddb9f98174187045bc580b072d

jmbrunskill:
> > Any new properties are added too if they're new but they're not updated. (To fix this we'd need to call the updateProperty graphql request in the dgraph - new issue to be created, but only if we want to allow updates...)
> 
> This makes sense - might be good to disable those existing property fields until we allow updates (if we do)?

I tried to disable them, but couldn't figureout how to get the state updating. [aa11e53](https://github.com/msupply-foundation/unified-codes/pull/489/commits/aa11e531c90fd4fb0c6a92a4aafa9ad82b8a317b)

@lache-melvin want to take a look?

I think we should probably allow editing in the longer term, I'm thing something like a lock icon you'd need to click and would unlock editing a field that would normally not be edited..

PR #491, opened 2024-01-04
490 empty fields get saved as new codes

### LINKED ISSUE #490: Empty fields get saved as new codes...
## Expected Behaviour
- [ ] UI should warn if something isn't filled in, and prevent saving
- [ ] Backend should skip nodes that have blank names? 

## Current Behaviour
image

jmbrunskill:
Fixes #490

## Description
Disables the Save button if you have a node created with no Name

image

image

image

Backend validates that there is a Name, Description, Type, and Category assigned.
It doesn't check length or anything else at this stage.


backend/service/src/universal_codes/upsert.rs
```diff
@@ -220,17 +224,18 @@ pub async fn validate(
...
         }
-        if new_entity.category.is_none() {
+        if new_entity.category.is_none()
+            || new_entity.category.clone().unwrap_or_default().is_empty()
+        {
             return Err(ModifyUniversalCodeError::InternalError(
                 "Category is required".to_string(),
             ));
         }
-        // TODO: check the type and category are valid
```
jmbrunskill, line 233: I'm not sure we need to do this??

lache-melvin (approved):
Sweet!

PR #492, opened 2024-01-05
487 audit logging for new or update codes

### LINKED ISSUE #487: Audit logging for new or update codes
## Expected Behaviour
The audit log should record what changes were made to a universal code, who made then, and 

## Current Behaviour
Logging isn't implemented because we can' pass context into an async function.

## Possible Solution
Not sure yet...

jmbrunskill:
Fixes #487 

## Description
Because the dgraph graphql library requires async functions, we weren't able t pass our normal `context` object around (with the storage connection it's not threadsafe)

To avoid this problem, I'm passing the user_id an service provider, then we can create connection in the function which needs to write to the audit log.
The storage connection is to allow functions to inherit database transactions for subsequent function calls.
Since we're not actually creating a transaction here, it's no needed.

Might be nice to create a new context struct though?

### DIFF
--- a/backend/graphql/universal_codes/src/mutations/upsert.rs
+++ b/backend/graphql/universal_codes/src/mutations/upsert.rs
@@ -28,7 +28,11 @@ pub async fn upsert_entity(
     match service_context
         .service_provider
         .universal_codes_service
-        .upsert_entity(input.into())
+        .upsert_entity(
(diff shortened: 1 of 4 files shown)


lache-melvin (approved):
Awesome James, great to have this in 🚀

PR #496, opened 2024-01-09
475 add vaccines

### LINKED ISSUE #475: Add a vaccine category
Vaccines are largely in the same hierarchy as drugs, but they have some additional complexity in that different vaccines for different strains, or maybe different brands, have different components.

Until now, this has been handled as a dose form qualification, but that's not really correct, because there's different ingredients and makeups. e.g. https://codes.msupply.foundation/detail/460d74bf

Vaccines either need to be their own category (or perhaps there is a flag on the drug that says it is a vaccine?) that adds this new element to our hierarchy.

Questions:
- Where in the hierarchy should this go?

lache-melvin:
Fixes #475 

## Description

Adds the vaccine category, and a filter for it on the browse page:
Screenshot 2024-01-09 at 3 21 55 PM

Changes `New Drug` in the app drawer to `New Item`, and places tabs along the top of the page to select what kind of item is being inputted. Is this obvious enough? Should it instead be some kind of initial please choose Drug/Consumable/Vaccine and that navigates you to the relevant form?
Screenshot 2024-01-09 at 3 27 58 PM

Adds a form to input new vaccines and edit them:
Screenshot 2024-01-09 at 3 21 13 PM

Displays vaccine data correctly - increases the depth of the entity query to show all info. Also adds the node type above the values, like you've mentioned before James.
Screenshot 2024-01-09 at 4 18 35 PM

Adds a new category node to the data loader code to create the Vaccine category, but I haven't added any actual vaccine data. Last I've seen, Louisa isn't quite done with the vaccine mapping yet, and her mappings don't include the universal codes. Once she's done though - I think the easiest way would be to create a new CSV for the vaccines to bring into the data loader? And leave the existing products.csv as is?


jmbrunskill:
Looks like we don't have a translation for PackSize (probably ok if we're cleaning it up)
image

jmbrunskill:
Should edit be working? it's not loading for me properly...

image


image

lache-melvin:
> Should edit be working? it's not loading for me properly...
> 
> image
> 
> image

Oh weird, yes mine is working... what is the value of the `type` property for the `Pneumococcal Vaccine` entity? It should be determining whether to render the drug for or the vaccine form depending on whether that type is `vaccine`...

frontend/system/src/Admin/EditDrug/helpers.ts
```diff
@@ -123,6 +124,164 @@ export const buildEntityFromDrugInput = (
...
+        .map(component => ({
+          ...getDetails(component),
+          brands:
+            component.children
+              ?.filter(brand => brand.type === 'brand')
+              .map(brand => ({
+                ...getDetails(brand),
+                routes:
+                  brand.children
+                    ?.filter(route => route.type === 'form_category') // form_category === route
```
jmbrunskill, line 144: // form_category === route

Since we have access to the loader code, should we just fix this in the data?
We probably need a mapping in the backend/filter to maintain backwards compatibility but might be a good plan?

jmbrunskill (approved):
Would be good to get this in and for some testing!

jmbrunskill:
> what is the value of the type property for the Pneumococcal Vaccine entity? Seems to be coming through as Unknown.
Maybe there's some backend issues?

image

I can investigate a bit further at my end.

jmbrunskill:
> Seems to be coming through as Unknown.

Fixed, I needed to run yarn load-data to add the Vaccine node into the hierarchy...

lache-melvin:
> > Seems to be coming through as Unknown.
> 
> Fixed, I needed to run yarn load-data to add the Vaccine node into the hierarchy...
> 
> I've put a backend fix to substitute the category in this situation which should be more robust...

Oh of course! Sorry I forgot to document that step 😅  Great fix - thank you!

frontend/system/src/Admin/EditDrug/helpers.ts
lache-melvin: Issue for sorting this: #500

PR #497, opened 2024-01-09
495 fix clickable whitespace

### LINKED ISSUE #495: Whitespace on item details page is clickable (links to edit page)
When logged in as an admin (so the `Update` button is available on the details page), the entire column of whitespace beneath the update button is also clickable, acting as the update button. 
 
See the hover/click cursor over the whitespace below!
Image

Maybe due to the button being wrapped by some kind of link element which has a full height?

lache-melvin:
Fixes #495

## Description

Switches from a Link component wrapper to using `navigate` on click. The Link component was taking on a full height, thus the clickable whitespace. We could have fixed this with CSS, but given the navigation is the result of the click, the navigate function seemed like the more appropriate option.

No longer clickable in the whitespace:
Screenshot 2024-01-09 at 4 34 22 PM

### DIFF
--- a/frontend/system/src/Entities/EntityDetails.tsx
+++ b/frontend/system/src/Entities/EntityDetails.tsx
@@ -6,12 +6,13 @@ import {
   ChevronDownIcon,
   EditIcon,
   Switch,
+  Box,
 } from '@common/ui';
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Looks good to me

PR #498, opened 2024-01-09
494 remove old source code

### LINKED ISSUE #494: Remove old source code
Now that the Rust API is connected and running, we should remove the remaining `data-service` folder, and its related scripts/README documentation - we should confirm the v1 API features are all in the Rust API before doing this.

We could also remove the `universal-codes` folder, which contains the source code for the old UI. I don't think we really need this anymore either?

lache-melvin:
Fixes #494

## Description

- Removes old web UI source code
- Remove `data-service` source code
- Moves TS integration tests from `data-service` dir to the frontend dir
- Removes documentation references to the old code


jmbrunskill (approved):
🚀

PR #502, opened 2024-01-10
500 remove api entity type mapping

### LINKED ISSUE #500: mSupply query doesn't receive new entities
Existing nodes (loaded in from the spreadsheet) have capitalised types (e.g. `Route/Form/DoseStrength/Unit`) - which is what our type filter searches DGraph for.

However, the API layer exposes a different set of types (e.g. `form_category/form/strength/unit_of_use`). These are the types we are setting in our upsert_entity mutations, but the backend isn't mapping them back to our capitalised DGraph types.

The web UI currently doesn't rely on the type filter, but the mSupply query does. It is searching for nodes of type `Strength` and `Unit`, but any new nodes have types `strength` or `unit_of_use`, so aren't returned in the query.

`upsert_entity` should map to existing DGraph node types, so that the mSupply query receives the new items too!

Or, if we don't see the mapping as particularly useful anymore, we support both!

lache-melvin:
Fixes #500

## Description

Removes the mapping to lowercase entity types in the API layer.

This will mean the type field in the response from the v1 api will change, but we're not aware of anything depending on those values, so it should fine?

Keeps the lower case variation in the filter in case there are existing queries using them - just having the uppercase ones pass through by default? Is that ok or should I create cases for them?


jmbrunskill (approved):
Looks good to me.

PR #504, opened 2024-01-11
499 Save and use Configuration data

### LINKED ISSUE #499: Ability to save configuration data
image

## Expected Behaviour
We have config screens for Routes, Forms, and Immediate Packaging.

Each should be able to be saved to either dgraph, or the sqlite db.
We should also query back this data and 

## Possible Solution
Saving this data to dgraph might make hierarchical data easier?
Although the relationships and hierarchies have defined levels so maybe it doesn't matter?

jmbrunskill:
Fixes #499

## Description
Save's configuration data for Route, Form, and Immediate Packaging to dgraph.
Configuration Options can be deleted
The new Drug form uses this data when populating options
Route, Form, and Immediate Packaging are derived from the existing data on `yarn load-data`

## Notes:
Run
`yarn load-data` and `yarn update-dgraph-schema` when testing this.


backend/dgraph/src/configuration/delete_configuration_item.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub data: UpsertResponse,
+}
+
+#[allow(non_snake_case)]
+#[derive(Deserialize, Debug, Clone)]
+pub struct UpsertResponse {
+    pub numUids: u32,
+}
+
+// Dgraph sometimes returns an error like this:
```
jmbrunskill, line 23: This actually happens quite a bit of you're deleting lots of items.
I think we should maybe put in somekind of random but very small delay between retries like between 1-5ms?

jmbrunskill: Arguably should also create some kind of helper function for this and/or see if we can set some kind of parameter to avoid these issues?

data-loader/data/v2/schema.graphql
```diff
@@ -22,3 +22,19 @@ type Property {
...
+type ConfigurationItem {
+  id: ID!
+  code: String! @id @dgraph(pred: "code") @search(by: [exact, trigram])
+  name: String @dgraph(pred: "name") @search(by: [exact, fulltext, trigram])
+  type: String @dgraph(pred: "type") @search(by: [exact])
+  children: [ConfigurationItem] @dgraph(pred: "subItems")
+  parents: [ConfigurationItem] @dgraph(pred: "~subItems")
+}
+
+type PropertyConfigurationItem {
```
jmbrunskill, line 35: This isn't implemented in this PR, if we want to make it configurable, we'll need to create another PR...

data-loader/src/v2/DataLoader.ts
```diff
@@ -117,6 +123,18 @@ export class DataLoader {
       } else {
         console.log(`WARNING: Failed to load entity with code ${entity.code}`);
       }
+      if (entity.type === 'Route') routes.add(entity.name.trim());
+      if (entity.type === 'Form') forms.add(entity.name.trim());
+      if (
+        entity.type === 'FormQualifier' &&
+        entity.name &&
+        entity.name.trim() &&
+        entity.name.trim().length < 20
```
jmbrunskill, line 132: This filters out a good bunch of the vaccines

frontend/system/src/Admin/Configuration/OptionEditModal.tsx
```diff
@@ -31,34 +30,44 @@ export const OptionEditModal = ({
...
-            // setErrorMessage(err.message);
-            onClose();
+            setIsLoading(true);
+            addEntity({
+              input: {
+                name: value,
+                type: type,
+              },
+            })
+              .catch(err => {
```
jmbrunskill, line 58: Should we disable this button for existing records and force a delete/insert?
Or should we allow editing/upsert?

lache-melvin: I could get behind add-only, though being able to fix typos is nice... if we do allow editing, might be good to rename the mutation to upsert rather than add?

Given changes to the entities are add-only as well I think that would be fine though!

Maybe rather than disabling the edit button once the modal is open, just remove the onclick on the table rows that opens the edit modal in the first place? At least until we have anything additional/interesting to display in the modal...

frontend/system/src/Admin/Configuration/OptionListConfigTab.tsx
```diff
@@ -17,46 +17,93 @@ import {
...
+      key: 'name',
       label: 'label.value',
     },
     'selection',
   ]);
 
-  const deleteAction = (rows: (ListOption | undefined)[]) => {
-    // TODO: implement delete
-    console.log('Rows to delete: ', rows);
+  const deleteAction = (rows: (ConfigurationItemFragment | undefined)[]) => {
```
lache-melvin, line 71: Just found a flaw in deleting these... while we store the values as strings against the entities, when we go to edit it, the dropdowns use the current config items as their options. See picture... I deleted the Oral option and now the disabled dropdown shows blank:

Screenshot 2024-01-12 at 10 15 40 AM

Perhaps in that edit form, we could append `entity.name` to the options list, so at least it displays correctly?

Otherwise I think we would be getting into storing some kind of soft delete flag to determine if its available as a selectable option or just a displayable option, which sounds like more work...

I think we'd run into the same issue if editing config item names was allowed too!

lache-melvin (approved):
Awesome work James!

Noticed you haven't pulled in develop since the Vaccines stuff was merged in - you'll need to bring in the config options into the vaccine edit form too 👍

data-loader/src/v2/DataLoader.ts
jmbrunskill: Yeah, probably over invested in this code. Nice to have something we can use to re-set the data though, so think it's worth having some kind of loader working...

data-loader/src/v2/DataLoader.ts
```diff
@@ -172,6 +190,92 @@ export class DataLoader {
...
+    }
+
+    // Create forms.
+    for await (const form of forms) {
+      if (!form) continue;
+      const nQuads = `
+        _:route <name> "${form}" .
+        _:route <code> "${form}" .
+        _:route <type> "Form" .
+        _:route <dgraph.type> "ConfigurationItem" .
```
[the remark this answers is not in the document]
jmbrunskill: [8984e68](https://github.com/msupply-foundation/unified-codes/pull/504/commits/8984e68ff6951f90213cdcee5391f0b12e80c13c)

frontend/system/src/Admin/Configuration/OptionListConfigTab.tsx
jmbrunskill: [8c4f82d](https://github.com/msupply-foundation/unified-codes/pull/504/commits/8c4f82d1af466c33d804b69c413af4ad92d1c710) I think this solves it quite nicely...

backend/dgraph/src/configuration/delete_configuration_item.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::DgraphClient;
+
+#[derive(Serialize, Debug, Clone)]
+struct DeleteVars {
+    code: String,
+}
+
+#[allow(non_snake_case)]
+#[derive(Deserialize, Debug, Clone)]
+pub struct UpsertResponseData {
```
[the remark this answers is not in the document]
jmbrunskill: [268bd1e](https://github.com/msupply-foundation/unified-codes/pull/504/commits/268bd1e276fa7a7ea57f1e7a6d4ec1088b29f3fa)

PR #505, opened 2024-01-11
480 pending changes

### LINKED ISSUE #480: UI for pending change details
Details view for a new product/changes to a product.

Figma details:

https://www.figma.com/file/LOK6QwbB4NQwyWYL68HQ6A/Universal-Codes?type=design&node-id=13-789&mode=design&t=WfiQGC3m8gJKnL7b-0#642671494

When we're adding to an existing drug, will we receive the existing dat a s well, or keep that separate and merge it somehow from the frontend??

lache-melvin:
Fixes #480

## Description

- Stores PendingChanges in Dgraph
- Displays these in the PendingChanges list view
- Creating/updating items create pending changes - no longer upserting

In an attempt not to make this PR any bigger, will create separate PRs to view the change details, and approve/reject them.

Went with storing the whole entity on the pendingChange body - felt significantly more complicated to store just change nodes and build the logic to integrate them, especially when it doesn't seem like updating existing items is going to be a particularly common thing to do? If I'm wrong then can definitely come back to this....

Am storing that updatedEntity as stringified JSON, because it doesn't look like DGraph supports nested JSON nodes, and storing a whole bunch of linked nodes just for the pending state seemed unnecessary? However, it would be good to have the logic like the duplication check in `upsert_entity` run when we create the change request, rather than when we approve it and run the upsert...

Should I send the entity to the backend as JSON and stringify it after validating? I assume that would mean we'd want to deserialise it before sending it back to the frontend... is that going to be painful to do with serde?


jmbrunskill (approved):
Looks great!

I'm keen to hold off on merging though as it will deploy to the test server till we have the review and approve done too?
Maybe just do a stacked PR?

PR #511, opened 2024-01-15
Use helper for dGraph retries

### LINKED ISSUE #510: Retry Helper for Dgraph
## Current Behaviour
Dgraph sometimes returns an error like this:
{"errors":[{"message":"Transaction has been aborted. Please retry","locations":[{"line":2,"column":3}],"extensions":{"code":"Error"}}],"data":null}
Seems to mostly happen in tests, but in case this happens in production we'll retry a few times
Sounds like this could possibly be to do with updating the same index? X-Dgraph-IgnoreIndexConflict?

## Expected Behaviour
Retries are in place, but there's a lot of duplicated code.
Some kind of helper code would be useful.

jmbrunskill:
Fixes #510 

## Description
Uses a helper function with re-tries to avoid mutation conflicts when doing parallel updates/deletes.

### Tests:
`cargo test --features=dgraph-tests`

### DIFF
--- a/backend/dgraph/Cargo.toml
+++ b/backend/dgraph/Cargo.toml
@@ -12,6 +12,7 @@ tokio = { version = "1", features = ["full"] }
 gql_client = "1.0.7"
 util = { path = "../util" }
 log = "0.4.14"
+rand = "0.8"
 
 [features]
 dgraph-tests = []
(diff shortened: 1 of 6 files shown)


lache-melvin (approved):
This is SUCH a nice refactoring thanks James! Makes the dgraph queries so much cleaner 🤩

PR #512, opened 2024-01-15
506 spamming the save button can result in duplicate records

### LINKED ISSUE #506: Spamming the save button can result in duplicate records
## Current Behaviour
image

## Possible Solution
Disable the button until the save process is complete.

1. Add a new records to an existing drug
2. Hit save lots of times

### LINKED ISSUE #507: Loading Indicator for Saving records
## Expected Behaviour
Some kind of loading indicator

## Current Behaviour
When using the save button on the Test server, it's not clear anything is happening, and feels broken...

1. Edit a drug
2. Click Save

## Notes
Saving feels slow on the server, might be worth investigating why?

jmbrunskill:
Fixes #506
Fixes #507

## Description
Adds a saving indicator while an entity is saving, this also avoids the issue with multiple button presses causing duplication.

### DIFF
--- a/frontend/system/src/Admin/EditDrug/DrugEditForm/DrugEditForm.tsx
+++ b/frontend/system/src/Admin/EditDrug/DrugEditForm/DrugEditForm.tsx
@@ -1,5 +1,11 @@
 import { useTranslation } from '@common/intl';
-import { Box, Typography, SaveIcon, ButtonWithIcon } from '@common/ui';
+import {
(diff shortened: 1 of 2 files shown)


lache-melvin (approved):
Nice one 🚀 Let's get this in - I agree with you that moving to useMutation might be even better, but I can make that refactor from the pending changes PR seeing as it has other mutations and touches this code too!

PR #513, opened 2024-01-15
480 pt2 view change details

### LINKED ISSUE #480: UI for pending change details

lache-melvin:
Fixes #480 

## Description

Adds the details page for a pending change:
- See diff of new changes
- Approve changes (creates/updates the entity)
- Reject changes
- Edit changes before approving

Still to do:
- Capturing `requestedFor`... this was on Louisa's mock up, i.e. `For: Fiji` - where should I capture this info? Like... a popup modal when you hit save?
- When reviewing a pending change, and go into edit mode, you can't tell which properties are existing and which are new. Would be good to get the existing properties disabled working, and perhaps also somehow highlighting the little button to open the properties modal - make it green? Something to indicate there are changes there... easily another PR!


backend/service/src/universal_codes/mod.rs
```diff
@@ -129,15 +136,33 @@ impl UniversalCodesService {
...
+        match result {
+            Some(pending_change) => Ok(Some(pending_change)),
+            None => Ok(None),
+        }
+    }
+
     pub async fn pending_changes(
         &self,
-        // filter: PendingChangesFilter, // TODO: need this once filtering by state?
+        // filter: PendingChangesFilter, // TODO: probably should expose this to allow filtering by state?
```
jmbrunskill, line 155: With this comment are you thinking something like see rejected changes, or see recently approved?
I don't think it's important at this stage, although I did think of the scenario where someone accidentally approves or accidentally rejects... I think that's another example of where we need the option to edit existing codes...

backend/service/src/universal_codes/tests/update_pending_change_body.rs
```diff
@@ -0,0 +1,190 @@
...
+                context.user_id.clone(),
+                request_id.clone(),
+                "new test body".to_string(),
+            )
+            .await;
+
+        assert!(result.is_err());
+
+        // TODO: A better way to remove new pending change from dgraph
+        // marking as rejected so as not to show up in PendingChange queries
```
jmbrunskill, line 181: Argueably this should be already rejected, but agree would be nice to have an actual delete.
With the data loader I'm finding this is kind of fine as we can just reset the database when ever we like as Devs, if we loose that ability I'd like to go trhough all the test data and make sure it's self cleaning or maybe we need to use a fresh dgraph namespace or something similar to having the individual sqlite databases.

frontend/system/src/Admin/PendingChanges/Details.tsx
```diff
@@ -0,0 +1,189 @@
...
+          ent.children?.forEach(addToExpandedIds);
+
+          ent?.properties && expandedIds.push(`${nodeId}_properties`);
+        }
+      };
+      addToExpandedIds(entity);
+
+      setExpanded(expandedIds);
+    }
+  }, [entity]);
```
jmbrunskill, line 75: Is it better to have 2 use effects rather than combining into a single one? Or is entity update somewhere else?

I don't have an issue with it how it is was just wondering, as it looks like could be combined with the one above...

frontend/system/src/Admin/EditDrug/DrugEditForm/DrugEditForm.tsx
```diff
@@ -137,11 +61,8 @@ export const DrugEditForm = ({
       },
     })
       .then(() => {
-        invalidateQueries();
-        setIsSaving(false);
-
         if (!initialEntity) {
-          // new entity - clear the form and show success snack
+          // new entity - clear the form
           success(
```
jmbrunskill, line 66: It feels like we should possibly navigate somewhere on submit, Feels like of weird to be staring at the same blank form after submitting.
Probably quite good from a UX though in terms of being able to quickly submit a bunch quickly, so on the fence...

frontend/common/src/intl/locales/en/system.json
```diff
@@ -31,6 +32,7 @@
   "label.add-route": "Add Route",
   "label.add-strength": "Add Strength",
   "label.add-unit": "Add Unit",
+  "label.approve-next": "Approve & Next",
   "label.brand": "Brand",
   "label.brands": "Brands",
   "label.component": "Comoponent",
```
jmbrunskill, line 38:
```suggestion
  "label.component": "Component",
```

Not from this PR, but noticed it here :)

jmbrunskill (commented):
This is probably more my problem than yours, however it looks like it's possible to enter duplicates still. I wonder if we need a frontend check and maybe a mutex in the backend to avoid this issue?
image

image

Let's create a new issue for this...

jmbrunskill (approved):
Looks great to me! Lets get it merged and tested!

frontend/system/src/Admin/EditDrug/DrugEditForm/DrugEditForm.tsx
lache-melvin: Yeah that was my thinking, will be good to hear what Louisa has to say, what the more common use case would be... possibly we could do some kind of Add/Add & Next buttons thing?

frontend/system/src/Admin/PendingChanges/Details.tsx
lache-melvin: Ah great catch - yeah the recommendation is to split up the useEffects with different dependencies, so the effect code only runs when it actually needs to... 

A previous iteration I had `entity` being updated more regularly, so separated that effect out from the one to run if the `pendingChange` changed... but now that code is gone, I've merged them back to one again (I think more efficient to queue all the state changes from one effect, so they are applied in one render) 

[1b9122f](https://github.com/msupply-foundation/unified-codes/pull/513/commits/1b9122f72c8106b859dec0c32c411f737d532d78)

backend/service/src/universal_codes/mod.rs
lache-melvin: Yeah that was my thinking, if we ever are needing to view recent changes - future thing for sure!

lache-melvin:
> This is probably more my problem than yours, however it looks like it's possible to enter duplicates still. I wonder if we need a frontend check and maybe a mutex in the backend to avoid this issue? image
> 
> image
> 
> Arguably the approve step should help avoid any one actually doing this I guess?

Ah damn - yeah can put a frontend check for duplicates in, should catch most of it. Yeah might have to rethink the parallelism in the backend, maybe we can run the validation over the whole entity rather than each node??

PR #519, opened 2024-01-16
515 update logo

### LINKED ISSUE #515: Update Logo
# Universal Codes Logo instead of mSupply Man

telegram-cloud-photo-size-1-4985733964778023958-x

lache-melvin:
Fixes #515

## Description

Adds UC logo to the app bar/drawer and favicon

Screenshot 2024-01-17 at 12 53 23 PM


frontend/host/src/components/Login/LoginIcon.tsx
```diff
@@ -1,21 +1,22 @@
...
   const [customLogo] = useLocalStorage('/theme/logo');
   const isSmallScreen = useIsSmallScreen() || small;
   const logoStyle = isSmallScreen
     ? { width: 61, height: 90 }
-    : { width: 122, height: 180 };
+    : { width: 285, height: 180 };
   const theme = useTheme();
 
-  if (!customLogo) return <MSupplyGuy style={logoStyle} />;
+  if (!customLogo)
```
jmbrunskill, line 18: We can remove the customLogo idea I think, but no real need too.

jmbrunskill (approved):
Looks good to me.

I do wonder if the text would render slightly nice if we used the font renderer rather than SVG Path?
We also should run it by Anisha...

lache-melvin:
> Looks good to me.
> 
> I do wonder if the text would render slightly nice if we used the font renderer rather than SVG Path? We also should run it by Anisha...

Oo I hadn't thought of rendering the font separately - do you think that would be worth it? I'll merge this so we can share with Anisha on the demo server...

PR #521, opened 2024-01-17
516 about page

### LINKED ISSUE #516: About page
## Expected Behaviour
The existing App has an about page, we should probably create something similar in our new app.

https://codes.msupply.foundation/about

lache-melvin:
Fixes #516

## Description

Adds about page:
Screenshot 2024-01-17 at 2 29 30 PM

Accessible both authenticated and unauthenticated... when authenticated, there is a link to About in the App Drawer. Have added nav links to the App Bar for when unauthenticated? Not sure how much I love this... could try adding the App Drawer back instead? With just the two links on it?


jmbrunskill (approved):
Looks great thanks!

PR #523, opened 2024-01-17
518 disable editing existing properties

### LINKED ISSUE #518: Disable editing for existing property
Image

Existing properties populate the property input fields in the edit modal, but property values aren't mutable. (Perhaps they should be one day, but for now, they're add-only.)

We should disable the fields with existing values, so the add-only nature is clear from the UI.

Also, the properties button changes from a `+` to a `>` to indicate that there are properties set for that node, but perhaps we could do something different to the `>` for when a node has existing properties to when there are new changes... make the button a different colour maybe??
Image

lache-melvin:
Fixes #518

## Description

Disables the input fields for properties that already exist:
Screenshot 2024-01-17 at 5 46 03 PM

We were setting the property code from the frontend - moved this to be done in the upsert (same time as when codes are set on new entities) so we can tell whether a property is new by whether or not it has a code.

Also changes the colour of the add properties buttons where there are changes to those properties:
Screenshot 2024-01-17 at 5 45 47 PM

- [ x] Tests OK.


jmbrunskill (approved):
Looks great, seems to work from a quick test

PR #525, opened 2024-01-17
Logo only navigates once

### LINKED ISSUE #447: Clicking logo (navigating to /browse) should only navigate once
## Expected Behaviour

Clicking the mSupply logo in the top left takes you back to the browse page. However, if you click it multiple times while on the /browse page, it shouldn't add any more entries to the history.

## Current Behaviour

If you click the logo 4 times, you'd have to click the back arrow 4 times to get back to wherever you were before 😅 

## Possible Solution

Kind of would have thought react-router would de-dupe stuff like this?

lache-melvin:
Fixes #447

## Description

Changes logo to be a link component so that it only navigates the user if we aren't already on that route. 

To test:
- While unauthenticated, navigate to a details page
- Click on logo to return to browse page
- Click the logo a bunch more times
- Click the back arrow in your browser 
- Result: You should only need to click the back arrow once to return to the details page


jmbrunskill (approved):
Ship it!

PR #530, opened 2024-01-18
526 new data loaders

### LINKED ISSUE #526: Create data loaders for vaccines and consumables
Louisa has mapped the existing Vaccine and Consumable data into their new structures, we now need to update the DataLoader to import these two new sheets.

lache-melvin:
Fixes #526

Using #536 as a base to avoid merge conflicts/rework - maybe review that first?

## Description

- Adds data loaders for newly mapped consumable/vaccine data
- Updates Vaccine hierarchy moving route and form up
- Updates msupply query filter to return new consumables types
- Removes vaccines and consumables from current product data csv

Note: consumables don't have properties... we discussed having the UNSPSC codes for the categories Louisa mapped the consumables to in an earlier iteration - shall I do this?


data-loader/src/v2/ConsumableDataParser.ts
```diff
@@ -0,0 +1,178 @@
...
+    };
+
+    try {
+      // Initialise duplicate codes.
+      const duplicates: string[] = [];
+
+      // Parse entity graph.
+      this.data.forEach(row => {
+        const productDefinition = [
+          // TODO: leaving this as Product for now... do we want that??
```
jmbrunskill, line 63: Yes I think so, as I'm using Product to find the top level of the tree to display when navigating to a lower level code

jmbrunskill (commented):
Looking good, I'll test it out once we get the import data.

backend/service/src/universal_codes/entity_filter.rs
```diff
@@ -21,7 +21,7 @@ pub fn dgraph_filter_from_v1_filter(filter: EntitySearchFilter) -> DgraphFilter
     // TODO: Better Search!
 
     // mSupply has a `special` way of querying that v1 API uses.
-    // and we need to set the type to ["Unit", "DoseStrength"] (in clause, where as normally this is a single string)
+    // and we need to set the type to ["Unit", "DoseStrength", "Presentation", "ExtraDescription"] (in clause, where as normally this is a single string)
```
jmbrunskill, line 24: BTW: I ran into this `single string` problem when trying to update mSupply.
I think we should probably make type accept an array unless that somehow breaks mSupply?

jmbrunskill: It is quite nice to be able to tweak the mSupply behaviour for the server, but it's a hacky method!

jmbrunskill (changes requested):
Looks good to me, have run the loader and spot checked stuff.

However I think we've ended up with a bunch of `1.04746E+16` type things again?

image

jmbrunskill:
```
wc -l data-loader/data/v2/drugs.csv 
    1134 data-loader/data/v2/drugs.csv
wc -l data-loader/data/v2/vaccines.csv
     126 data-loader/data/v2/vaccines.csv
wc -l data-loader/data/v2/consumables.csv 
      67 data-loader/data/v2/consumables.csv
```
= 1327 rows

```
wc -l data-loader/data/v2/products_old.csv 
    1327 data-loader/data/v2/products_old.csv
```

jmbrunskill (approved):
Good to go!

PR #535, opened 2024-01-22
components -> active ingredients

lache-melvin:
Fixes #528

## Description

Renames `component` to `activeIngredient` - makes this renaming both in the translations and in code/for the entity node type, seeing as no vaccine nodes exist yet.

Screenshot 2024-01-22 at 12 53 13 PM

PR #538, opened 2024-01-22
536 vaccine name details

### LINKED ISSUE #536: Add optional Name Details field to vaccines
Some vaccines have a friendlier name that they are known by e.g. for the influenza vaccine: `Southern Hemisphere 2021, Egg-based`

This should be an optional field at the same level as active ingredients

lache-melvin:
Fixes #536

Uses #535 as a base to avoid too many merge conflicts - maybe review that one first!

## Description

Adds optional Vaccine Name Details Section:
Screenshot 2024-01-22 at 5 03 22 PM


jmbrunskill (approved):
Looks good to me, gave it a bit of a test.

It feels like it would be nice to have a way to add the level into an existing hierarchy.
Like if we start with Just `Influenza`  then later decide to split it to Seasonal variations.
I don't think that's super straightforward unfortunately but maybe worth creating a new issue for it?

lache-melvin:
> Looks good to me, gave it a bit of a test.
> 
> It feels like it would be nice to have a way to add the level into an existing hierarchy. Like if we start with Just `Influenza` then later decide to split it to Seasonal variations. I don't think that's super straightforward unfortunately but maybe worth creating a new issue for it?

Oo yeah that would be cool, though agree would be a bit more work at this time - I'll create the issue though!

PR #539, opened 2024-01-22
517 remove requested for

### LINKED ISSUE #517: Capture requestedFor for pending changes

lache-melvin:
Fixes #517

## Description

Removes the `requestedFor` field from pending changes:
Screenshot 2024-01-22 at 5 20 51 PM

Removes the property entirely, rather than just not displaying... this field was mostly to explain _why_ the change is being requested, which we'll only really need once external submissions are available, but something like a `reason` field will probably make more sense then.


jmbrunskill (approved):
🚀

PR #540, opened 2024-01-22
redirect old details route to new page

### LINKED ISSUE #483: Existing links should work when we deploy new UC version
The existing universal codes app has 2 routes:

- https://codes.msupply.foundation
- https://codes.msupply.foundation/detail/{code}

The new app currently has these pages on:
- /browse
- /browse/{code}

If anyone has links to the existing UC pages, they should still work. Either we should move the browse pages to match the old paths, or add some redirects so the users end up in the right place :)

P.S. We do have a redirect from `/` to `/browse`, so it's just the details page we need to worry about! That one ends up as a 404

lache-melvin:
Fixes #483

## Description

Redirects `/details/:code` pages to `/browse/:code`, so links from the old web app will get redirected as expected in the new web app!

Test:
- Go to: http://localhost:4201/details/c7750265
- You should be redirected to http://localhost:4201/browse/c7750265 and see the Abacavir details

### DIFF
--- a/frontend/host/src/RedirectDetails.tsx
+++ b/frontend/host/src/RedirectDetails.tsx
@@ -1,0 +1,17 @@
+import { RouteBuilder } from '@common/utils';
+import { AppRoute } from 'frontend/config/src';
+import React from 'react';
+import { Navigate, useParams } from 'react-router-dom';
+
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
🎉

PR #541, opened 2024-01-22
509 suggest existing code

### LINKED ISSUE #509: Suggest existing code when entering a new drug
## Expected Behaviour
If a description for a drug (and potentially other items) looks like it could be a duplicate, the user should get a warning, and ideally a link to the right edit form
image

## Notes
Saw this with user testing, seems natural to enter one drug variant, the going new drug again to add the next, which returns an error.

lache-melvin:
Fixes #509

## Description

Adds a `Did you mean...` suggester to the new item forms, which links to the edit page for that item:
Screenshot 2024-01-23 at 8 56 00 AM

Not sure whether to debounce the display of the suggestion a little?

### DIFF
--- a/frontend/common/src/intl/locales/en/system.json
+++ b/frontend/common/src/intl/locales/en/system.json
@@ -86,6 +86,7 @@
   "property-who_eml": "WHO EML Categories",
   "property-code_unspsc": "UNSPSC Code",
   "message.status-updated": "Change request {{ status }} for {{ name }}",
+  "message.did-you-mean": "Did you mean {{ suggestion }}?",
   "message.entity-created": "Request submitted to create {{ name }}",
   "message.entity-updated": "Request submitted to update {{ name }}",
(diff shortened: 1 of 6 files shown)


jmbrunskill (approved):
Looks good to me, I wonder if a fuzzy match would be better, but does have the problem of false positives, so this is a good starting point.

I think a debounce is worth it, I tried typing in with a high latency internet connection throttle ended up with heaps of inflight http requests.
image

lache-melvin:
> Looks good to me, I wonder if a fuzzy match would be better, but does have the problem of false positives, so this is a good starting point.
> 
> I think a debounce is worth it, I tried typing in with a high latency internet connection throttle ended up with heaps of inflight http requests. 

Yeah I wondered about the fuzzy match, as it would be nice to catch a slight typo etc., but yep felt the exact match was more reliable at this point - will be good to hear what the feedback is though! 

Debounce added 👍

PR #543, opened 2024-01-22
520 duplication issues

### LINKED ISSUE #520: Duplication issues
- [ ]               Weirdly I seem to be able to create duplicates of existing items (e.g. from the Loader)
image
However, after that I get an error if I try to do it again, I can't see anything like a space showing in the description would would explain not receiving the duplication error...

image

_Originally posted by @jmbrunskill in https://github.com/msupply-foundation/unified-codes/issues/476#issuecomment-1894684311_

- [ ]               There's no warning about duplicated nodes in the frontend
- [ ] If loading in a entity with internal duplication, the duplication check doesn't notice it and creates the duplicated nodes

lache-melvin:
Fixes #520

## Description

- Fixes backend logic checking for duplicates (was checking for more than 1 duplicate)
- Shows error in the frontend if creating duplicate nodes (not checking the product name, but I think the `Did you mean` thing in #541 would help cover that possibility?)
- Disables submit button if there are duplicate nodes

Screenshot 2024-01-23 at 12 18 30 PM


jmbrunskill (approved):
Love it, the siblings thing seems to work really well.

PR #544, opened 2024-01-22
531 only mark change approved if upsert ok

### LINKED ISSUE #531: Bug: Pending change "Not in pending state" after upsert failure
To reproduce:

- Add new drug with same name as an existing drug (do this twice bc of the bug where it doesn't seem to pick up the duplication the first time...)
- Click Approve the pending change
- See the duplication error
- Click Reject on the pending change
- BUG: Reject should work - instead get an error that the change is no longer in a pending state

Would appear that the "approved" status is still getting applied even if the upsert fails (and we then can't reject an approved change.) Should only get to that step if the upsert goes through smoothly.

lache-melvin:
Fixes #531

## Description

Checks upsert completed successfully before marking a Pending Change as Approved

### Tests:

- [X] Tests OK.


jmbrunskill (approved):
Nice, quite handy that having a missing name triggers the error state for the test.
Good comment there to explain it.

PR #554, opened 2024-01-25
537 imm pack pack size in edit forms

### LINKED ISSUE #537: Add immediatePackaging to drug edit form
Immediate packagings are displayed in browse mode:

Image

But not available in edit mode:
Image

I _think_ we agreed we don't need to be able to input immediate packaging anymore, but perhaps even if we don't put the `add` button, we should display the existing ones!

This also goes for existing pack sizes. For both, there are some where previous levels of the hierarchy have been skipped. This is not something we want to support going forward, but we should deal with these existing ones.

lache-melvin:
Fixes #537

## Description

Adds immediate packaging and Pack Size editing to Drug Edit form, and Pack Size editing to Vaccine edit form:
Screenshot 2024-01-26 at 12 10 37 PM


jmbrunskill (approved):
Looks good.
Snuck in a fix for the capitalisation of `VIal` as I noticed it when testing this PR...

PR #561, opened 2024-01-29
558 allow querying property config items when unauthenticated

### LINKED ISSUE #558: Property config values aren't available when not logged in!
There's an admin privileges check for viewing the property config item values, but they're need to display the drug data:

Image

We should make reading the config items publicly available, just mutating them for admins! (whoops 🙈)

lache-melvin:
Fixes #558 

## Description

Removes MutateUniversalCodes auth check on queries for property config items, so they are queryable when unauthenticated:

Screenshot 2024-01-30 at 10 37 14 AM


jmbrunskill (approved):
Good find!

PR #562, opened 2024-01-29
551 duplicate names in dataloaded data

### LINKED ISSUE #551: Duplicate Names in DataLoaded Data
image

jmbrunskill:
Fixes #551 

## Description
Two products had the same product name but different codes, this seems to be the case all the way back to 4.3 version spreadsheets.
I've updated to use one (effectively deleting a code, but unlikely to be being used anywhere)

Wrote a quick python script to check for others product name code miss matches, but seems ok.


lache-melvin (approved):
🚀

PR #564, opened 2024-01-31
559 GS1 codes

### LINKED ISSUE #548: Collapse nodes lower than the unit level
Currently we expand all nodes of the entity trees when displaying them on the details page. However, for drugs and vaccines, typically we'd only need to see down to the unit level.

We should collapse immediatePackaging and packSizes where they are present.

### LINKED ISSUE #559: Support GS1 Codes
Implement MVP support for GS1 Codes...:

- Add button to entity details (and hide nodes lower than unit?):
Image

- Table display view for GS1s:
Image

- An input page to assign a GS1 to an entity

lache-melvin:
Fixes #559 
Fixes  #548 

## Description

Adds add, view and delete capability for GS1 codes, with the restriction that the related PackSize node already exists.

View all GS1 codes:
Screenshot 2024-02-01 at 9 18 27 AM

Sorting the list e.g. by manufacturer is currently not supported, and searching is also not implemented. Unfortunately,  I don't think its possible for us to sort or search the whole GS1 list by product, as DGraph requires those fields to be scalar types... if that search capability is desired in future, we could duplicate the entity data onto the GS1 node to be able to search on it.

Add modal searches existing pack size entities:
Screenshot 2024-02-01 at 9 19 04 AM

See number of related GS1s (is related a better word than linked? also GS1s or GTINs?) for a unit - lower nodes collapsed by default:
Screenshot 2024-02-01 at 9 19 45 AM

Lower nodes expandable if you want them to be:
Screenshot 2024-02-01 at 9 20 39 AM

Link takes you to a filtered view of GS1s for only that product:
Screenshot 2024-02-01 at 9 22 00 AM

Add modal PackSize options are filtered accordingly:
Screenshot 2024-02-01 at 9 23 46 AM

### Tests:

- [x] Tests OK.


lache-melvin:
Support for viewing barcodes when unauthenticated added!

backend/service/src/barcodes/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        first: Option<u32>,
+        offset: Option<u32>,
+    ) -> Result<BarcodeCollection, BarcodeServiceError> {
+        let result = barcodes(&self.client, BarcodeQueryVars { first, offset })
+            .await
+            .map_err(|e| BarcodeServiceError::InternalError(e.message().to_string()))?; // TODO: Improve error handling?
+
+        match result {
+            Some(data) => Ok(BarcodeCollection {
+                total_length: usize_to_u32(data.data.len()),
```
jmbrunskill, line 95: Probably should use the actual total/count from dgraph here since you have the first/offset, it won't return the right value when you have a limit on the results less than the total available records...

backend/service/src/barcodes/upsert.rs
```diff
@@ -0,0 +1,117 @@
...
+    new_barcode: &AddBarcode,
+) -> Result<(), ModifyBarcodeError> {
+    if new_barcode.gtin.is_empty() {
+        return Err(ModifyBarcodeError::InternalError(
+            "GTIN is required".to_string(),
+        ));
+    }
+
+    if new_barcode.manufacturer.is_empty() {
+        return Err(ModifyBarcodeError::InternalError(
```
jmbrunskill, line 91:
```suggestion
        return Err(ModifyBarcodeError::BadUserInput(
```
This is more a user error, similar for the others in `validate`

data-loader/data/v2/schema.graphql
```diff
@@ -74,6 +75,13 @@ type PropertyConfigurationItem {
   url: String @dgraph(pred: "url")
 }
 
+type Barcode {
+  id: ID!
+  gtin: String! @id @dgraph(pred: "code") @search(by: [exact, trigram])
```
jmbrunskill, line 80: I like this using `code` in dgraph, kind of as our `id` look up.
Have created `interaction_group_id` and `interaction_id` in my other PR, but definitely complicates things a little bit...

frontend/common/src/ui/components/toolbars/DeleteLinesDropdownItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useNotification } from '@common/hooks';
+import { useTranslation } from '@common/intl';
+import { DeleteIcon, DropdownMenuItem } from '@common/ui';
+import {
+  LocalStorage,
+  RecordWithId,
+  useConfirmationModal,
+} from 'frontend/common/src';
+
+export const DeleteLinesDropdownItem = <T extends RecordWithId>({
```
jmbrunskill, line 11: Nice! Like this component!

frontend/host/src/Site.tsx
```diff
@@ -92,7 +92,7 @@ export const Site: FC = () => {
                   />
                 }
               />
-              <Route path={'/details/:code'} element={<RedirectDetails />} />
+              <Route path={'/detail/:code'} element={<RedirectDetails />} />
```
jmbrunskill, line 95: Was this deliberate? I can't remember what the old link was like...

frontend/system/src/Entities/BarcodeLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const getBarcodeCount = (entity: EntityData) => {
+  let count = 0;
+
+  count += entity.barcodes.length;
+
+  const countChildBarcodes = (e: EntityData) => {
+    e.children?.forEach(c => {
+      count += c.barcodes.length;
+      countChildBarcodes(c);
```
jmbrunskill, line 48: Interesting that you can call a function you're creating as a const inside the function by the name. I guess it makes sense, but not obvious to me...

jmbrunskill (commented):
Looks awesome, heaps of work!

I tried to add a barcode and got this error

```
 Internal error: Failed to parse response: Error("invalid type: null, expected struct Entity", line: 1, column: 6409). The response body is: {"errors":[{"message":"Non-nullable field 'code' (type String!) was not present in result from Dgraph.  GraphQL error propagation triggered.","locations":[{"line":3,"column":3}],"path":
```

Will try to investigate further tomorrow to see what's gone wrong.

lache-melvin:
> Looks awesome, heaps of work!
> 
> I tried to add a barcode and got this error
> 
> ```
>  Internal error: Failed to parse response: Error("invalid type: null, expected struct Entity", line: 1, column: 6409). The response body is: {"errors":[{"message":"Non-nullable field 'code' (type String!) was not present in result from Dgraph.  GraphQL error propagation triggered.","locations":[{"line":3,"column":3}],"path":
> ```
> 
> Will try to investigate further tomorrow to see what's gone wrong.

Oh I had this issue yesterday viewing drug details on the 'loading in standard config' branch! I think I fixed it - have just merged develop in... if you know which product you tried to create the barcode for, would you mind trying again?

jmbrunskill (approved):
Looks great! Couple of comments but merge away when you're ready.

frontend/host/src/Site.tsx
lache-melvin: Oh whoops didn't realise that made it in here... when we deployed the production version i realised the old links were /detail/:code, but we'd set up the redirect off /details... will do via separate PR for clarity 😎

frontend/system/src/Entities/BarcodeLink.tsx
lache-melvin: Yeah... I suppose because it is defined above where it is called it all works out ok? It is a little funky though!

backend/service/src/barcodes/mod.rs
lache-melvin: [df8a34d](https://github.com/msupply-foundation/unified-codes/pull/564/commits/df8a34db49193b3f931935ba57acd6268e084635)

backend/service/src/barcodes/upsert.rs
lache-melvin: [428c3ca](https://github.com/msupply-foundation/unified-codes/pull/564/commits/428c3cae2ae490823d967b05f69e86836b5a2405)

PR #566, opened 2024-02-01
fix entity update button clickability

### LINKED ISSUE #565: Entity update button is hard to click
The tooltip for the entity update button blocks your ability to click the button (in the area that the tooltip covers)...

lache-melvin:
Fixes #565 

## Description

Moves the tooltip component inside the button component, so the button is still fully clickable even when the tooltip renders.


jmbrunskill (approved):
🚀

PR #569, opened 2024-02-01
fix the detail redirect

### LINKED ISSUE #483: Existing links should work when we deploy new UC version

lache-melvin:
Fixes #483

## Description

Submitted a PR to fix the details page redirect last week but I redirected from /details instead of /detail 🙈

### DIFF
--- a/frontend/host/src/Site.tsx
+++ b/frontend/host/src/Site.tsx
@@ -92,7 +92,7 @@ export const Site: FC = () => {
                   />
                 }
               />
-              <Route path={'/details/:code'} element={<RedirectDetails />} />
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
⛴ it

PR #570, opened 2024-02-01
446 store browsing categories in URL

### LINKED ISSUE #446: Store browsing filters in URL params
## Expected Behaviour

If you're on the browse page, and you change the filter (default is `drug`, so either selected both `drugs` and `consumables` or just `consumables`) it would be nice for this filter state to persist if you refresh, or go to a details page and then hit the back arrow.

## Current Behaviour

Going back to the browse page sets the filter back to the default state of only showing drugs.

## Possible Solution

Store the filter state in the URL, so that it persists through refreshing/navigating to and from the browse page.

lache-melvin:
Fixes #446

## Description

Manages category filters as url query params so they persist on refresh/navigating back:
Screenshot 2024-02-02 at 11 54 08 AM

### DIFF
--- a/frontend/system/src/Entities/ListView.tsx
+++ b/frontend/system/src/Entities/ListView.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { useTranslation } from '@common/intl';
 import {
   AppBarContentPortal,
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Feels much better!

PR #573, opened 2024-02-02
571 duplicate description

### LINKED ISSUE #571: Duplicate description/key error for artemether/lumefantrine
There is a duplicated description Artemether/Lumefantrine Oral Tablet (dispersible) 20mg/120mg that should be consolidated. 

Image

lache-melvin:
Fixes #571 

## Description

Removes the form qualifiers from being loaded in by the data loader.

Updates the fuzzy search autocomplete to use `code` for the label rather than description, as we have done for other autocomplete components (though both code and description should be unique...)

So I found a lot of duplicates... there are a few cases:

- Some resulted from merging Form and Form Qualifier. Nodes had the same Form code, but now that the qualifier is merged in, their form values are actually different. These are safe to resolve as we can use the old qualifier code as the new form code.
- There were also a couple of copy paste errors with the vaccine sheet that created duplicate codes. Went back to the old codes to resolve these so I think these are fine to change.
- Some had a unique code where they shouldn't, creating duplicate descriptions - we'll lose codes resolving these, but leaving them breaks the ability to edit/add to that product, because the upsert code checks for duplicates...


jmbrunskill (approved):
Good finds!

PR #594, opened 2024-02-08
584 include reference in drug interaction query

### LINKED ISSUE #584: Drug interaction reference box not saving
## Current Behaviour
Reference box not saving information

https://github.com/msupply-foundation/unified-codes/assets/143057869/f88bb768-145d-4264-ad85-a4daab900c24

## Expected Behaviour
Save information entered

lache-melvin:
Fixes #584

## Description

Includes reference in the dgraph query so we see the results of what we saved 😁 

https://github.com/msupply-foundation/unified-codes/assets/55115239/1a30f452-5561-4bd4-b84d-7d1e5f01260f

## Note

We've had a few of these issues (or at least I've bumped up against them devving...) especially when adding new fields to existing types, there being no type warning between the DGraph queries and the resulting rust struct... wonder if there is anything we could do for this going forward?

### DIFF
--- a/backend/dgraph/src/drug_interaction/interactions.rs
+++ b/backend/dgraph/src/drug_interaction/interactions.rs
@@ -23,6 +23,7 @@ query drugInteractions($name: String) {
     description
     action
     severity
+    reference
     drugs {
         name
         description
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Seems like a good idea to include this hahaha

PR #595, opened 2024-02-08
583 add consumable packsizes barcodes

### LINKED ISSUE #583: Consumables unable to add barcode
## Current Behaviour
Barcode not available for consumable items as pack size is not in the consumable mapping

## Possible Solution
Add pack size option to adding consumables at the end as optional so that it can have barcodes added

lache-melvin:
Fixes #583

## Description

Adds pack sizes the consumables - allowing the addition of barcodes:

Adding pack sizes in the consumable edit form - had a brief wondering if we would want immediate packaging before pack size here (i.e. box, bag...) but Louisa doesn't think this extra layer of data to input would be worth it at this stage... which I guess means it will need to be optional if we do ever introduce it in future...
Screenshot 2024-02-09 at 11 45 40 AM

Once approved, those pack sizes are available in the barcodes input modal:
Screenshot 2024-02-09 at 11 46 07 AM

Pack sizes hidden by default:
Screenshot 2024-02-09 at 11 46 28 AM

Expanded:
Screenshot 2024-02-09 at 11 46 33 AM

Pack size view of barcodes:
Screenshot 2024-02-09 at 11 46 42 AM

### Tests:

- [x] Tests have been updated to cover changes introduced by this PR.

### DIFF
--- a/frontend/system/src/Admin/EditEntity/helpers.test.ts
+++ b/frontend/system/src/Admin/EditEntity/helpers.test.ts
@@ -9,19 +9,25 @@ import {
   isValidDrugInput,
   isValidVaccineInput,
 } from './helpers';
-import { ConsumableInput, DrugInput, VaccineInput } from './types';
+import {
+  ConsumableInput,
+  DrugInput,
+  EntityDetails,
+  VaccineInput,
+} from './types';
 
 describe('getAllEntityCodes', () => {
   it('returns empty array when entity is not defined', () => {
     expect(getAllEntityCodes(undefined)).toMatchObject([]);
   });
   it('returns array of codes with empty strings filtered out', () => {
-    const entityDetails = {
+    const entityDetails: EntityDetails = {
       code: '7c8c2b5b',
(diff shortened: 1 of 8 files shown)


frontend/system/src/Admin/EditEntity/helpers.test.ts
```diff
@@ -9,26 +9,33 @@ import {
...
   });
   it('returns array of codes with empty strings filtered out', () => {
-    const entityDetails = {
+    const entityDetails: EntityDetails = {
       code: '7c8c2b5b',
       name: 'Acetic Acid',
       type: 'drug',
       alternativeNames: [],
       properties: [],
+      barcodes: [],
```
lache-melvin, line 30: these type issues should have been caught earlier... I think this is another instance of where not running TS compilation before tests, nor ensuring that the tests themselves compile, is a little unfortunate

jmbrunskill (approved):
Looks good to me!

Sorry for the delay reviewing!

PR #606, opened 2024-02-16
Add properties to entities query

### LINKED ISSUE #605: Properties aren't coming through to mSupply
[This needs to be filled out!!]

There appears to be a bug where properties are no longer included in our API response to the mSupply query...

jmbrunskill:
Fixes #605 

## Description
The dgraph backend `entities` query was deliberately kept minimal to avoid having huge dgraph requests just to get product names, however the mSupply query is asking for properties.

```
query{
	entities(filter: { description: "para" type: "drug unit_of_use medicinal_product" orderBy: { field: "description" descending: false } } offset: 0 first: 25) { data { code description type properties { type value } product { properties { type value } } }, totalLength }
}
```

Now returns:

```
{
  "data": {
    "entities": {
      "data": [
        {
          "code": "bbfcf518",
          "description": "Paracetamol Oral Suspension 24mg per mL",
          "type": "DoseStrength",
          "properties": [],
          "product": null
        },
```

## Checklist:
-[ ] Also need to make product properties work, as they should be included in mSupply too...

### Tests:

- [x] Tests OK.

### DIFF
--- a/backend/dgraph/src/entities.rs
+++ b/backend/dgraph/src/entities.rs
@@ -60,6 +60,12 @@ query EntitiesQuery($filter: EntityFilter, $first: Int, $offset: Int, $order: En
     code
     alternative_names
     __typename
+    properties {
+        __typename
+        code
+        type
+        value
+    }
     parents {
         id
         name
@@ -125,6 +131,7 @@ mod tests {
         assert_eq!(data.data[0].description, "Heparin Sodium");
         assert_eq!(data.data[0].r#type, "Product");
         assert_eq!(data.aggregates.unwrap().count, 1);
+        assert!(!data.properties.is_empty())
     }
 
     #[tokio::test]


lache-melvin (approved):
Oh nice one - lets definitely get this in for mSupply!

It's leading me to wonder if we should consider creating a new stripped down query for our UI though? Or maybe just flag as something to load test in future... we do this entities query for all items for the fuzzy search - especially as we add more items I can imagine this query getting slow?

lache-melvin:
> Also need to make product properties work, as they should be included in mSupply too...

Oh good call - do you mean that product field is used in mSupply? Are you intending to include that in this PR as well or a separate piece of work?

jmbrunskill:
> we do this entities query for all items for the fuzzy search - especially as we add more items I can imagine this query getting slow?

Yeah, it definitely doesn't feel great. Could compare speed to using a loader for all properties? Still seems fast locally on good hardware, but I agree! Maybe an issue for the future?

jmbrunskill:
> Oh good call - do you mean that product field is used in mSupply? Are you intending to include that in this PR as well or a separate piece of work?

Was going to do this in the same PR, but since it's approved, I'll merge this and do the other in a new one.

PR #609, opened 2024-02-20
Fix the Product request on entities query

### LINKED ISSUE #608: Fix `Product` field in v1 API
## Expected Behaviour
```
{ "query": " { entities(filter: { description: \\"para\\" type: \\"drug unit_of_use medicinal_product\\" orderBy: { field: \\"description\\" descending: false } } offset: 0 first: 25) { data { code description type properties { type value } product { properties { type value } } }, totalLength } } " }
```

Should return properties for the associated `Product`, usually the Drug level properties that are more likely to be included.

## Current Behaviour
Product is null

## Possible Solution
GraphQL Loader

jmbrunskill:
Fixes #608

## Description
Requesting `product` on an `entities` query such as mSupply does now works.
```
query{
	entities(filter: { description: "para" type: "drug unit_of_use medicinal_product" orderBy: { field: "description" descending: false } } offset: 0 first: 25) { data { code description type properties { type value } product { properties { type value } } }, totalLength }
}
```

### Tests:

No tests have been added, would be nice to have those integration test we had in jest to update...

### DIFF
--- a/backend/graphql/core/Cargo.toml
+++ b/backend/graphql/core/Cargo.toml
@@ -12,6 +12,7 @@ doctest = false
 repository = { path = "../../repository" }
 service = { path = "../../service" }
 util = { path = "../../util" }
+graphql_v1_core = { path = "../../graphql_v1/core" }
 
 actix-web = { version = "4.0.1", default-features = false, features = ["macros", "cookies"] }
 anymap = "0.12"
@@ -24,4 +25,4 @@ serde = "1.0.126"
 serde_json = "1.0.66"
 thiserror = "1.0.30"
 tokio = { version = "1.29", features = ["macros" ] }
-async-std = "1.12"
+async-std = { version = "1", features = ["attributes", "tokio1"] }
--- a/backend/graphql/core/src/lib.rs
+++ b/backend/graphql/core/src/lib.rs
@@ -5,6 +5,8 @@ pub mod simple_generic_errors;
 pub mod standard_graphql_error;
 pub mod test_helpers;
 
+use graphql_v1_core::loader::LoaderRegistry as LoaderRegistryV1;
+
 use std::sync::Arc;
 
 use actix_web::cookie::Cookie;
@@ -34,6 +36,7 @@ pub trait SelfRequest: Send + Sync {
 pub trait ContextExt {
     fn get_connection_manager(&self) -> &StorageConnectionManager;
     fn get_loader<T: anymap::any::Any + Send + Sync>(&self) -> &T;
+    fn get_loader_v1<T: anymap::any::Any + Send + Sync>(&self) -> &T;
     fn service_provider(&self) -> Arc<ServiceProvider>;
     fn service_context(
         &self,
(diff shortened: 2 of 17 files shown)


backend/graphql/core/Cargo.toml
```diff
@@ -24,4 +25,4 @@ serde = "1.0.126"
 serde_json = "1.0.66"
 thiserror = "1.0.30"
 tokio = { version = "1.29", features = ["macros" ] }
-async-std = "1.12"
+async-std = { version = "1", features = ["attributes", "tokio1"] }
```
jmbrunskill, line 28: This cost me a tonne of time yesterday.

I think because of the need to call async functions in dgraph from with async graphql, this `tokio1` feature is needed but was hard to find!

backend/graphql/lib.rs
```diff
@@ -76,6 +77,7 @@ pub fn schema_builder() -> Builder {
 pub fn build_schema(
     connection_manager: Data<StorageConnectionManager>,
     loader_registry: Data<LoaderRegistry>,
+    loader_registry_v1: Data<LoaderRegistryV1>,
```
jmbrunskill, line 80: I went with 2 loader registries, but does feel a bit messy having the /v1/ api.
Tempted to suggest we move everything to the main one and just have the /v1/ point to the same thing?
Would be for future...

lache-melvin:
> No tests have been added, would be nice to have those integration test we had in jest to update...

We do have these? https://github.com/msupply-foundation/unified-codes/blob/develop/frontend/integration/v1_graphql_entity.int.ts 

Run with `yarn test-integration`

jmbrunskill:
> > No tests have been added, would be nice to have those integration test we had in jest to update...
> 
> We do have these? https://github.com/msupply-foundation/unified-codes/blob/develop/frontend/integration/v1_graphql_entity.int.ts
> 
> Run with `yarn test-integration`

Ah nice, for some reason I thought we'd lost them as we were going to move to rust ones. I'll update...

backend/graphql/core/src/lib.rs
```diff
@@ -34,6 +36,7 @@ pub trait SelfRequest: Send + Sync {
 pub trait ContextExt {
     fn get_connection_manager(&self) -> &StorageConnectionManager;
     fn get_loader<T: anymap::any::Any + Send + Sync>(&self) -> &T;
+    fn get_loader_v1<T: anymap::any::Any + Send + Sync>(&self) -> &T;
```
lache-melvin: Just for my understanding... how come you're defining the v1 loader here on the graphql/core ContextExt as well? From what I can see it's only used from the graphql_v1/core ContextExt??

backend/graphql/lib.rs
lache-melvin: Yeah I agree, rather than the main one having to point to v1 for the entity queries! Definitely a future refactor 😎

lache-melvin (approved):
Working great! Thanks James 🙏

lache-melvin:
Actually! Tests aren't compiling... 😅 

```
error[E0609]: no field `properties` on type `EntityData`
   --> dgraph/src/entities.rs:134:23
    |
134 |         assert!(!data.properties.is_empty())
    |                       ^^^^^^^^^^ unknown field
    |
    = note: available fields are: `data`, `aggregates`
```

PR #617, opened 2024-03-21
Remove parallel upserts for child entities

### LINKED ISSUE #616: when approving drug it saved itself but does not appear in the drug list and says it already exists
IMAGE 2024-03-21 10:17:53

Entered in diclofenac drug details to be submitted into system. On the approval process when i clicked approve it came up with big blob of text in a red box with warning - tried screenshotting this but it vanished. 

my approval box did not change however so i thought it was not approved yet, clicked approve but this time says the drugs already have existing codes as below:

IMAGE 2024-03-21 10:18:03

tried searching for the drugs in the search bar and it comes up but nothing comes up to select in the big box area: 

IMAGE 2024-03-21 10:18:15

jmbrunskill:
Fixes #616 

## Description
Discovered that parallel child node inserts seems to trigger dgraph transaction errors.
For some reason this was not seen with earlier inserts...

I've increased the retry time to 20ms, so hopefully increases the chances for the dgraph system to recover before retries kick in.

### Tests:
- [X] No new tests required

I've tested adding Diclofenac as well as editing it.
I'm slightly concerned there was a reason I did children first, before parent but seems to work fine in this order...

### DIFF
--- a/backend/service/src/universal_codes/upsert.rs
+++ b/backend/service/src/universal_codes/upsert.rs
@@ -55,47 +55,37 @@ pub async fn upsert_entity(
         },
     };
 
-    let mut child_handles = vec![];
-    // First process any children
-    if let Some(children) = updated_entity.children.clone() {
-        for child in children {
-            // Using spawn_local partly to avoid errors with async recursion
-            // https://docs.rs/async-recursion/latest/async_recursion/ would be another approach, but this also gets us some parallelism
-            let result = tokio::task::spawn_local(upsert_entity(
-                sp.clone(),
-                user_id.clone(),
-                client.clone(),
-                child,
-            ));
-            child_handles.push(result);
-        }
-    }
-
-    // Wait for the children to finish
-    // Collect the codes of any children, as their codes might be generated in the recursion
-    let mut child_codes = vec![];
-    for handle in child_handles {
-        let result = handle.await;
-        let child_entity = match result {
-            Ok(child_entity) => child_entity?,
-            Err(e) => {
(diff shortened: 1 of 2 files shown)


backend/service/src/universal_codes/upsert.rs
```diff
@@ -55,47 +55,37 @@ pub async fn upsert_entity(
...
+            let child_entity = handle.await.map_err(|e| {
+                ModifyUniversalCodeError::InternalError(format!(
+                    "Join Error with spawn_local upsert_entity: {:?}",
                     e
-                )))
-            }
-        };
-        child_codes.push(child_entity.code);
+                ))
+            })??;
```
jmbrunskill, line 85: Is double `??` ok, or better to be explicit?

jmbrunskill:
We still have an issue where a pending change might not be insertable in future, however I think it should now at least create the parent element first, so worst case we can reject a broken pending change an re-enter the other records.

It would be possible for all entities to be inserted by their links could be missing. Not sure if there's a good way to handle without a transction.
We could create our own transaction of sorts where we try to undo but that sounds pretty complex...

backend/service/src/universal_codes/upsert.rs
lache-melvin: Fine to me? If you're down in the weeds enough to care about one `?`, you can probably notice two?

lache-melvin (approved):
> I'm slightly concerned there was a reason I did children first, before parent but seems to work fine in this order...

Yeah I had the same thought... but working fine for me too - let's get it in! 🤞

